PR #3, opened 2019-10-28
Fix infinite loop

chrismwendt:
Prior to this change, we'd enter an infinite loop when files used symbols from each other.

After this change, all definitions are visited before references, eliminating the possibility of an infinite loop.

### DIFF
--- a/src/main/java/DocumentIndexer.java
+++ b/src/main/java/DocumentIndexer.java
@@ -22,7 +22,6 @@ public class DocumentIndexer {
     private Emitter emitter;
     private Map<String, DocumentIndexer> indexers;
     private String documentId;
-    private boolean indexed = false;
     private Set<String> rangeIds = new HashSet<>();
     private Map<Range, DefinitionMeta> definitions = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


src/main/java/ProjectIndexer.java
```diff
@@ -58,7 +58,15 @@ public void index() {
...
-            indexer.index();
+            indexer.preIndex();
+        }
+
+        for (DocumentIndexer indexer : indexers.values()) {
+            indexer.visitDefinitions();
+        }
+
+        for (DocumentIndexer indexer : indexers.values()) {
+            indexer.visitReferences();
```
chrismwendt: I split up `index()` into 3 pieces:

- `preIndex()` emits some metadata about the doc (similar in function to `postIndex()`)
- `visitDefinitions()`
- `visitReferences()`

src/main/java/DocumentIndexer.java
```diff
@@ -106,23 +103,26 @@ private void linkUses(DefinitionMeta meta, String documentId) {
         }
     }
 
-    private class LSIFVisitor extends CtScanner {
-        // TODO add support for more language constructs:
-        // https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/CtScanner.java
+    // TODO add support for more language constructs:
+    // https://github.com/INRIA/spoon/blob/master/src/main/java/spoon/reflect/visitor/CtScanner.java
 
+    private class DefinitionsVisitor extends CtScanner {
```
chrismwendt: I split `LSIFVisitor` into 2 pieces:

- `DefinitionsVisitor`
- `ReferencesVisitor`

Most of the diff in this part of the PR is caused by me moving all the reference nodes to `ReferencesVisitor` and definition nodes to `DefinitionsVisitor`.

src/main/java/DocumentIndexer.java
```diff
@@ -106,23 +103,26 @@ private void linkUses(DefinitionMeta meta, String documentId) {
...
         @Override
         public <T> void visitCtParameter(CtParameter<T> el) {
+            super.visitCtParameter(el);
```
chrismwendt: I previously forgot a few of these `super` calls.

PR #18, opened 2020-03-12
emit required toolInfo

rvantonder:
Tried to upload a dump file that worked after I added tool name.

```
File: dump.lsif
Root: 
error: 500 Internal Server Error

{"message":"Could not find tool type in metadata vertex at the start of the dump."}
```

It'd be nice to introduce tests so that we can make sure our indexers are up to spec if stronger server validation is introduced.


efritz:
I've been pushing for this class of tests since my first week....

rvantonder:
> I've been pushing for this class of tests since my first week....

We'll get there.

PR #31, opened 2020-08-27
Fixed project root absolute path ending in '.' when lsif-java called with -projectRoot .

Strum355:
Must please the almighty sourcegraph/lsif-validate :pray:

### DIFF
--- a/src/main/java/ProjectIndexer.java
+++ b/src/main/java/ProjectIndexer.java
@@ -26,11 +26,11 @@ public class ProjectIndexer {
         this.emitter = emitter;
     }
 
-    public void index() {
+    public void index() throws IOException {
         emitter.emitVertex("metaData", Util.mapOf(
                 "version", "0.4.0",
                 "positionEncoding", "utf-16",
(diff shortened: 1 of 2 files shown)


src/main/java/ProjectIndexer.java
```diff
@@ -27,10 +27,17 @@ public ProjectIndexer(Arguments arguments, Emitter emitter) {
...
+		Path projectRootPath = Paths.get(arguments.projectRoot).toAbsolutePath();
+		String projectRoot = projectRootPath.toString();
+		if (projectRootPath.getFileName().toString().equals(".")) {
+			projectRoot = projectRoot.substring(0, projectRoot.lastIndexOf('.'));
+		}
+
         emitter.emitVertex("metaData", Util.mapOf(
                 "version", "0.4.0",
                 "positionEncoding", "utf-16",
-                "projectRoot", String.format("file://%s", Paths.get(arguments.projectRoot).toAbsolutePath().toString()),
```
efritz, line 33: If you just use `getCanonicalPath` instead does it still have the same issue?

Strum355: Good catch, that works well :+1:

PR #36, opened 2020-09-07
Updated README section on generating maven project from gradle

Strum355:
For non multi-project Gradle projects, it's assumed that the `java` plugin is used. The updated version of this, compared to previously, will handle the case where `sourceSets.main.java.srcDirs` is non-default
For multi-project Gradle projects, more work is involved to generate `<modules>` blocks

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,6 @@ publishing {
                     node.appendNode("build").with {
                         if (sourceSets.size() > 0) {
                             appendNode("sourceDirectory", sourceSets.first())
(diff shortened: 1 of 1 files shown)


efritz (approved):

Strum355:
Please note, that this is a far from all-inclusive solution. There are a lot of scenarios in which these instructions do not work, most obviously Android projects (yet), where sourceSets.main.java.srcDirs.size() > 1 and Im sure many others

PR #47, opened 2020-10-01
Bump spoon version

### LINKED ISSUE #44: lsif-java on master no longer indexes Spoon
As of 2de844423a3d94075b845b9d18b30b883b8b967c a commit introduced a regression that causes `lsif-java` to spin forever (>5 hours on GH action) when indexing the Spoon repository. [Example](https://github.com/INRIA/spoon/runs/1142710290). I've reverted the lsif-java-action to use the last known good commit for indexing Spoon, since it currently uses `lsif-java` in their CI: https://github.com/sourcegraph/lsif-java-action/commit/11dd91b4cba0f2a143914b14162d27d202f6726e. Context from [spoon thread](https://github.com/INRIA/spoon/pull/3394#issuecomment-698196728).

Since lsif-java is under development, I won't harp on this, but what is the current thinking on testing (to catch/avoid regressions) and in future, support stable versions (so we can revert to safe commit versions, in case of regressions)?

Strum355:
Should close #44 (untested)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ repositories {
 
 dependencies {
     compile 'com.github.javaparser:javaparser-symbol-solver-core:3.16.1'
-    compile 'fr.inria.gforge.spoon:spoon-core:8.3.0-beta-8'
+    compile 'fr.inria.gforge.spoon:spoon-core:8.3.0-beta-10'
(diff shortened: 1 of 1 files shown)


rvantonder (approved):
Got nothing to lose, so approving :-) Thanks! I'll confirm later

PR #63, opened 2020-10-20
First iteration of gradle: fetching of source directories and classpath

Strum355:
Importing a gradle project as an Eclipse natively with the gradle tooling API seems to be...the best way to go about this? Will follow up with multi-project support and investigations later

whack


efritz:
:smoke:

efritz (approved):

PR #67, opened 2020-10-27
Quote project root path in replace regex

vilunov:
Fixes handling of windows-style paths with backslashes


BullDog57Rus (approved):

Strum355 (approved):
Thanks for the PR! I dont have a windows machine to test on but this seems straightforward enough to simply accept : )

PR #85, opened 2021-02-11
Implement a new design based on a Java compiler plugin that generates SemanticDB

olafurpg:
This PR removes most of the contents from the original repo and introduces a new architecture for lsif-java.

In a nutshell, the new implementation includes a [Java compiler plugin](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.compiler/com/sun/source/util/Plugin.html) that generates one [SemanticDB](https://scalameta.org/docs/semanticdb/specification.html) file for every `*.java` source file. After compilation completes, the SemanticDB files are processed to produce LSIF.

semanticdb-javac-pipeline(1)

The updated readme includes more details about the benefits of this new design.

You can try out the new precise code intelligence from the link below. Observe that navigating to a method name `app()` goes to the precise implementation even if there are many classes that implement methods with the same name `app()`.

https://sourcegraph.com/github.com/sourcegraph/lsif-java@3cab1b8/-/blob/tests/minimized/src/main/java/minimized/MinimizedJavaMain.java#L13

Screenshot 2021-02-11 at 12 28 16

There are still several known bugs like the incorrect highlighting in this "Find references" request

Screenshot 2021-02-11 at 12 29 59

The combined diff in this PR is large, but the Java compiler plugin implementation is only 745 lines of code if you exclude the generated protobuf definitions. The bulk of the diff comes from the new ["snapshot testing"](https://jestjs.io/docs/en/snapshot-testing) infrastructure to index a corpus of the https://github.com/airbnb/epoxy library.


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/MD5.java
```diff
@@ -0,0 +1,35 @@
...
+    char[] hexChars = new char[bytes.length * 2];
+    int j = 0;
+    while (j < bytes.length) {
+      int v = bytes[j] & 0xFF;
+      hexChars[j * 2] = HEX_ARRAY[v >>> 4];
+      hexChars[j * 2 + 1] = HEX_ARRAY[v & 0x0F];
+      j += 1;
+    }
+    return new String(hexChars);
+  }
```
Strum355, line 26: I _think_ this could be replaced with [javax.xml.bind.DataTypeConverter.printHexBinary](https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/DatatypeConverter.html#printHexBinary-byte:A-) but no big deal

Strum355 (approved):
Nothing obvious stands out, LGTM :shipit: good stuff :+1:

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/MD5.java
olafurpg: IntelliJ says this class exists in `rt.jar` but I can't find it in my Java home with

Screenshot 2021-02-11 at 14 10 11

```
❯ jar tf jre/lib/rt.jar  | grep javax.xml.bind.DataTypeConverter
<empty>
❯ javap -cp jre/lib/rt.jar javax.xml.bind.DataTypeConverter
Error: class not found: javax.xml.bind.DataTypeConverter
❯ javap -cp lib/tools.jar javax.xml.bind.DataTypeConverter
Error: class not found: javax.xml.bind.DataTypeConverter
```

PR #86, opened 2021-02-15
Add benchmark results

olafurpg:
The benchmark results indicate that the overhead from the compiler plugin is ~45% compared to normal compilation. Single-core compilation performance is ~28k lines of code per second (excluding comments).


olafurpg:
I found this related discussion on jvmti error 23 from the author of async-profiler https://twitter.com/AndreiPangin/status/1050168444912656384 My understanding is that the overhead from the method calls marked as "jvmti error 23" is precise but we don't know the names of those methods because they have been unloaded. 

Screenshot 2021-02-15 at 14 01 44

I suspect the large performance overhead comes from classloading the compiler plugin for every compilation unit. It's worth trying to see if we can cache the compiler instance for multiple source files.

Strum355 (approved):

olafurpg:
I force pushed a new commit that fixes a performance bug in how we run the Java compiler in the tests. The overhead is now "only" 45-50% compared to normal compilation, which is significantly faster than the initial 4x slowdown that the benchmarks original showed.

PR #91, opened 2021-02-25
JDK 9+ Support

### LINKED ISSUE #90: JDK 11+ support for javac plugin indexer
In Olaf's testing with graalvm JDK 11, there were compile issues with certain symbols from the com.sun.source packages not being resolved. 
To support constructs beyond JDK 8 (records etc) we'll need to be able to compile with a newer JDK.

This was previously achieved in the Kotlin rewrite through compiling with runtime java version checks to avoid calling methods that don't exist in earlier versions.
Another approach we could try is with [multi-release jars](https://www.baeldung.com/java-multi-release-jar).

The issue of tools.jar not being a thing in JDK9+ is already taken care of by Olaf in the [SBT plugin](https://sourcegraph.com/github.com/sourcegraph/lsif-java@main/-/blob/project/JavacBootClasspathPlugin.scala)

Strum355:
Previously, the semanticdb-javac compiler plugin only supported Java 8. The plugin failed when compiling with Java 9+ due to our use of 'unsupported' APIs that changed from JDK 9.

This PR adds support for all JDKs from 9 onwards in addition to the previously supported JDK 8 by 
1) Using APIs that have been stable (thus far) and 
2) Removing usages of (as tested thus far) redundant and unstable APIs.

Notes:
- With JDK9+ anonymous classes and everything enclosed within don't get skipped. We should see why this happens in JDK8
- There was 1 erroneous reference that moved to 0:0, why? https://github.com/sourcegraph/lsif-java/pull/91#discussion_r583017177

Closes #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,11 @@ These are the main components of the project.
 | `./sbt`                                                            | terminal | Start interactive sbt shell with Java 8. Takes a while to load on the first run.    |
 | `unit/test`                                                        | sbt      | Run fast unit tests.                                                                |
 | `~unit/test`                                                       | sbt      | Start watch mode to run tests on file save, good for local edit-and-test workflows. |
-| `snapshot/testOnly tests.MinimizedSnapshotSuite`                   | sbt      | Runs fast snapshot tests. Indexes a small set of files under `tests/minimized`.     |
-| `snapshot/testOnly tests.MinimizedSnapshotSuite -- *InnerClasses*` | sbt      | Runs only individual tests cases matching the name "InnerClasses".                  |
-| `snapshot/testOnly tests.LibrarySnapshotSuite`                     | sbt      | Runs slow snapshot tests. Indexes a corpus of external Java libraries.              |
-| `snapshot/test`                                                    | sbt      | Runs all snapshot tests.                                                            |
-| `snapshot/run`                                                     | sbt      | Update snapshot tests. Use this command after you have fixed a bug.                 |
+| `snapshots/testOnly tests.MinimizedSnapshotSuite`                   | sbt      | Runs fast snapshot tests. Indexes a small set of files under `tests/minimized`.     |
+| `snapshots/testOnly tests.MinimizedSnapshotSuite -- *InnerClasses*` | sbt      | Runs only individual tests cases matching the name "InnerClasses".                  |
+| `snapshots/testOnly tests.LibrarySnapshotSuite`                     | sbt      | Runs slow snapshot tests. Indexes a corpus of external Java libraries.              |
+| `snapshots/test`                                                    | sbt      | Runs all snapshot tests.                                                            |
+| `snapshots/run`                                                     | sbt      | Update snapshot tests. Use this command after you have fixed a bug.                 |
 | `fixAll`                                                           | sbt      | Run Scalafmt, Scalafix and Javafmt on all sources. Run this before opening a PR.    |
 
 ### Import the project into IntelliJ
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
@@ -1,12 +1,10 @@
 package com.sourcegraph.semanticdb_javac;
 
-import com.sun.tools.javac.code.Kinds;
-import com.sun.tools.javac.code.Scope;
 import com.sun.tools.javac.code.Symbol;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.IdentityHashMap;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import java.util.*;
 
 import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
 
@@ -33,7 +31,7 @@ public final class GlobalSymbolsCache {
   }
 
   public boolean isNone(Symbol sym) {
-    return sym == null || sym.kind == Kinds.NIL || (sym.kind & Kinds.ERRONEOUS) != 0;
+    return sym == null;
   }
 
   private String uncachedSemanticdbSymbol(Symbol sym, LocalSymbolsCache locals) {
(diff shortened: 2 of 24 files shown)


tests/snapshots/src/main/generated/minimized/Enums.java
```diff
@@ -1,4 +1,5 @@
 package minimized;
+^^^^^ reference minimized/Enums#
```
Strum355, line 2: This one is unusual :thinking: @olafurpg

README.md
```diff
@@ -105,11 +105,11 @@ These are the main components of the project.
...
 | `~unit/test`                                                       | sbt      | Start watch mode to run tests on file save, good for local edit-and-test workflows. |
-| `snapshot/testOnly tests.MinimizedSnapshotSuite`                   | sbt      | Runs fast snapshot tests. Indexes a small set of files under `tests/minimized`.     |
-| `snapshot/testOnly tests.MinimizedSnapshotSuite -- *InnerClasses*` | sbt      | Runs only individual tests cases matching the name "InnerClasses".                  |
-| `snapshot/testOnly tests.LibrarySnapshotSuite`                     | sbt      | Runs slow snapshot tests. Indexes a corpus of external Java libraries.              |
-| `snapshot/test`                                                    | sbt      | Runs all snapshot tests.                                                            |
-| `snapshot/run`                                                     | sbt      | Update snapshot tests. Use this command after you have fixed a bug.                 |
+| `snapshots/testOnly tests.MinimizedSnapshotSuite`                   | sbt      | Runs fast snapshot tests. Indexes a small set of files under `tests/minimized`.     |
```
olafurpg, line 108: 👍

tests/snapshots/src/main/generated/com/airbnb/epoxy/EpoxyAsyncUtil.java
```diff
@@ -148,6 +149,7 @@ public static Handler createHandler(Looper looper, boolean async) {
     }
 
     return new Handler(looper);
+//         ^^^^^^^^^^^^^^^^^^^ reference `<init>`#
```
olafurpg, line 152: This looks like an improvement 👍

tests/snapshots/src/main/generated/com/airbnb/epoxy/EpoxyControllerAdapter.java
```diff
@@ -581,21 +581,53 @@ void notifyModelChanged(int position) {
...
       new ItemCallback<EpoxyModel<?>>() {
+//    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ reference `<any>`#`<init>`# 15:7
 //        ^^^^^^^^^^^^ reference _root_/
+//        ^^^^^^^^^^^^ reference _root_/
+//                     ^^^^^^^^^^ reference com/airbnb/epoxy/EpoxyModel#
 //                     ^^^^^^^^^^ reference com/airbnb/epoxy/EpoxyModel#
         @Override
+//       ^^^^^^^^ reference java/lang/Override#
         public boolean areItemsTheSame(EpoxyModel<?> oldItem, EpoxyModel<?> newItem) {
+//                     ^^^^^^^^^^^^^^^ definition com/airbnb/epoxy/EpoxyControllerAdapter#ITEM_CALLBACK.``#areItemsTheSame().
```
olafurpg, line 592: Improvement 👍

tests/snapshots/src/main/generated/minimized/Enums.java
olafurpg: 🤔 Hmmm not sure what's going on here. It's not a blocking issue. I think it makes sense that we try to document these quirks at some later point as issues. For now, it's most important to the the infrastructure working.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
```diff
@@ -97,18 +104,13 @@ private String uncachedSemanticdbSymbol(Symbol sym, LocalSymbolsCache locals) {
    * SemanticDB spec</a>.
    */
   private String methodDisambiguator(Symbol.MethodSymbol sym) {
-    Scope.Entry lookup =
-        sym.owner.members().lookup(sym.name, s -> s instanceof Symbol.MethodSymbol);
-    ArrayList<Symbol> peers = new ArrayList<>();
-    while (lookup != null) {
-      if (lookup.sym != null) {
-        peers.add(lookup.sym);
+    Iterable<? extends Element> elements = sym.owner.getEnclosedElements();
```
olafurpg, line 107: Nice find 👍 This looks like a decent replacement for `members().lookup()`

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
```diff
@@ -32,12 +35,16 @@ public String semanticdbSymbol(Symbol sym, LocalSymbolsCache locals) {
     return result;
   }
 
+  public boolean isNone(TypeMirror mirror) {
+    return mirror == null || mirror.getKind() == TypeKind.NONE;
+  }
+
   public boolean isNone(Symbol sym) {
-    return sym == null || sym.kind == Kinds.NIL || (sym.kind & Kinds.ERRONEOUS) != 0;
+    return sym == null; // || sym.kind == Kinds.Kind.NIL || sym.kind == Kinds.Kind.ERR;
```
olafurpg, line 43: This seems to be working ok. No? I'm OK with removing the comment.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
```diff
@@ -32,12 +35,16 @@ public String semanticdbSymbol(Symbol sym, LocalSymbolsCache locals) {
...
   }
 
   private String uncachedSemanticdbSymbol(Symbol sym, LocalSymbolsCache locals) {
-    if (isNone(sym)) return SemanticdbSymbols.NONE;
+    if (sym == null) return SemanticdbSymbols.NONE;
```
olafurpg, line 47:
```suggestion
    if (isNone(sym)) return SemanticdbSymbols.NONE;
```

olafurpg (changes requested):
The diffs look like a mix of nice improvements and minor regressions. I think it's OK to merge these changes and follow up with a PR that adds new testing infrastructure to compile with java 8 and java 11 within the same sbt build.

Some blocking issues:

* Update CI configuration to use java 11. 
```diff
+++ .github/workflows/ci.yml
      - uses: olafurpg/setup-scala@v10
+       with:
+         java-version: adopt@1.11
```
* Update readme to reflect Java 11 support.
* Update title and description of the PR.The format I try to follow in PR description is something like "Previously, the semanticdb-javac compiler plugin only supported Java 8. The plugin crashed when compiling with Java 11. This PR adds support for Java 11 so that the plugin works on both Java 8 and Java 11.". 
* Manually confirm that the plugin work on Java 8. We can fix the testing infrastructure later to automatically prevent regressions.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/GlobalSymbolsCache.java
Strum355: The commented part is not backwards compatible. JDK 8 it was `Kinds.NIL` etc, JDK 9+ the enum was moved to be a nested enum. The behaviour appears to be identical at best, but I'm unsure if we're testing enough edge cases to surface where this change may not be sufficient

tests/snapshots/src/main/generated/com/airbnb/epoxy/EpoxyAsyncUtil.java
Strum355: Im assuming this isnt correct yet, but progress?

olafurpg (approved):
LGTM 🚀

PR #100, opened 2021-03-01
Add lsif-java command-line tool with Gradle and Maven support.

### LINKED ISSUE #87: Gradle support

olafurpg:
Previously, users needed to add manual configuration to their
Gradle/Maven builds in order to use lsif-java. Now, users can run the
new `lsif-java` command-line tool to automatically produce LSIF.

In order to automatically produce LSIF, `lsif-java` needs to compile all
the Java sources with an additional Java compiler options
`-Xplugin:semanticdb`. Most build tools support a way to add custom
compiler options but that typically requires updating build
configuration files such as `pom.xml` for Maven and `build.gradle` for
Gradle. However, we want to avoid updating build configuration files.

The exact solution to update the compilation options varies between
build tools:

* Maven: you can customize the `javac` executable through the
  command-line flag `mvn -Dmaven.compiler.executable=PATH compile`
* Gradle: you can pass in a custom Gradle script via
  `gradle --init-script=PATH`, which allows you to query information out
  of the build and configure settings. By default, `lsif-java` will try
  to configure the build to use a custom `javac` script similar to
  Maven. However, this approach does not work for Gradle builds that use
  the the newly released "Java toolchains" feature.  For such builds, we
  can provide a custom `java` binary that registers a custom Java agent
  that ensures that Gradle always enabled the required
  `-Xplugin:semanticdb` flag.

Fixes #87 
Fixes #86


README.md
```diff
@@ -13,17 +13,24 @@ Currently, only Java 8 with the build tool sbt is supported. We hope to increase
...
-| Java 11+         | ❌      |
+| Language version | Support            |
+| ---------------- | ------------------ |
+| Java 7           | ❌                 |
+| Java 8           | ✅                 |
+| Java 11          | ✅                 |
+| Java 12          | Untested, may work |
+| Java 13          | Untested, may work |
+| Java 14          | Untested, may work |
+| Java 15          | Untested, may work |
```
Strum355, line 24: Works on Java 15, just not formally in the CI test, so Im assuming youre leaving it out for that reason?

cli/src/main/scala/com/sourcegraph/lsif_java/buildtools/GradleBuildTool.scala
```diff
@@ -0,0 +1,103 @@
...
+        List(
+          gradleCommand,
+          "--init-script",
+          script,
+          "-Porg.gradle.java.installations.auto-detect=false",
+          "-Porg.gradle.java.installations.auto-download=false",
+          s"-Porg.gradle.java.installations.paths=${toolchains.paths()}",
+          s"--no-daemon"
+        )
+      buildCommand ++= index.finalBuildCommand(List("clean", "compileTestJava"))
```
Strum355, line 55: Curious what this is for, why `compileTestJava` is hardcoded

README.md
olafurpg: Strictly speaking, the "Java 8" row should have the same message as Java 15. The next item on my agenda is to get cross-java CI working so this is only temporary.

cli/src/main/scala/com/sourcegraph/lsif_java/buildtools/GradleBuildTool.scala
olafurpg: Is there a better default command for gradle? I tried `-x test build` but it often skips compilation of test files and it also triggers integration tests. The user can override this command like this
```
lsif-java -- my custom gradle command
```

Strum355: Ah okay didn't know its overridable, seems fine then :slightly_smiling_face: Ive a feeling some of the android projects arent emitting semanticdb files due to using different compile task names, but Im not sure, would've thought `build` would catch it

olafurpg: The injected gradle settings specifically target the java plugin, which isn't used in android projects. I saw that the errorprone plugin had android specific logic, which we may need to learn from

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/InjectSemanticdbOptions.java
```diff
@@ -0,0 +1,28 @@
+package com.sourcegraph.semanticdb_javac;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
+public class InjectSemanticdbOptions {
```
Strum355, line 9: Just for the sake of someone coming to read this in the future, it mightnt be a bad idea to briefly explain how this works. Its a fairly dense function :smile:

Strum355 (approved):
Nothing glaring stands out, great work on this :+1: Time for me to really learn Scala now lol

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/InjectSemanticdbOptions.java
olafurpg: Good catch! This is the hairiest part of the PR and it was definitely missing documentation. I was considering writing this logic in a bash script originally, but figured it would be easier to maintain as a small Java program.

PR #120, opened 2021-03-09
fixes snapshot generator comment // insertion

Strum355:
You can see the change this has by looking at the committed snapshot files :slightly_smiling_face: 

For cases where there is no 1 or more whitespace of leeway, we cut down the number of `^` by one. This mostly affects type decls, who's ranges need to be fixed _anyways_ and annotations on decls

### DIFF
--- a/tests/unit/src/main/scala/tests/SemanticdbPrinters.scala
+++ b/tests/unit/src/main/scala/tests/SemanticdbPrinters.scala
@@ -49,12 +49,17 @@ object SemanticdbPrinters {
       }
     out
       .append(
-        if (r.getStartCharacter > 3)
+        if (r.getStartCharacter > 2)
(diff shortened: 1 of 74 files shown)


tests/unit/src/main/scala/tests/SemanticdbPrinters.scala
```diff
@@ -49,12 +49,17 @@ object SemanticdbPrinters {
       }
     out
       .append(
-        if (r.getStartCharacter > 3)
+        if (r.getStartCharacter > 2)
           "// " + " " * (r.getStartCharacter - 3)
         else
-          " " * r.getStartCharacter
+          "//" //* r.getStartCharacter
```
olafurpg, line 55:
```suggestion
          "//"
```

olafurpg (approved):
LGTM 👍

PR #121, opened 2021-03-10
fixes range deriving for constructs incl classes/interfaces

### LINKED ISSUE #109: Incorrect range for interface definition
https://github.com/sourcegraph/lsif-java/blob/756563ffca0a55d49141fc2c94d75af729c30744/tests/snapshots/src/main/generated/minimized/Interfaces.java#L3-L4

```diff
-- obtained
++ expected
  public interface Interfaces {
- //     ^^^^^^^^^^ definition minimized/Interfaces#
+ //               ^^^^^^^^^^ definition minimized/Interfaces#

```

Strum355:
Uses text search derived from previous com.sun.source lsif-java (Kotlin edition), which is originally derived from georgewfraser's java language server. Closes #109 

Of particular note is that the current changes cause definition occurence to [not be emitted](https://github.com/sourcegraph/lsif-java/pull/121/files#diff-1c43dc4d39a78f8bfeee7dbff8bd285a2cd2c711e57d8053586575921be9230cL325). The reference to this no longer existing definition occurrence still gets emitted, we can either follow up with a PR or fix it in this PR. 

I feel like moving parts of `SemanticdbVisitor.semanticdbRange(...)` to the `RangeFinder` class to keep more of the logic together, but I will revisit this tomorrow.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
@@ -8,11 +8,16 @@ import com.sun.source.util.Trees;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.Name;
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
+
 public class RangeFinder {
+  private static final HashMap<CompilationUnitTree, String> contentsCache = new HashMap<>();
+
   public static Optional<Semanticdb.Range> findRange(
       TreePath path, Trees trees, CompilationUnitTree root, Element element, int startPos) {
     LineMap lineMap = root.getLineMap();
@@ -22,11 +27,11 @@ public class RangeFinder {
     int endPos = (int) trees.getSourcePositions().getEndPosition(root, path.getLeaf());
     // false for anonymous classes
     if (name.length() != 0) {
-      startPos = findNameIn(root, name, startPos, endPos);
+      startPos = findNameIn(root, name, startPos);
       endPos = startPos + name.length();
     }
 
-    if (endPos == -1) {
+    if (endPos == -1 || startPos == -1) {
       return Optional.empty();
     }
 
@@ -40,19 +45,21 @@ public class RangeFinder {
     return Optional.of(range);
   }
 
-  private static int findNameIn(CompilationUnitTree root, CharSequence name, int start, int end) {
-    String contents;
-    try {
-      contents = root.getSourceFile().getCharContent(true).toString();
-    } catch (IOException e) {
-      return -1;
-    }
+  private static int findNameIn(CompilationUnitTree root, CharSequence name, int start) {
+    String content =
+        contentsCache.computeIfAbsent(
+            root,
+            (s) -> {
+              try {
+                return root.getSourceFile().getCharContent(true).toString();
+              } catch (IOException e) {
+                return null;
+              }
+            });
 
-    if (end == -1) {
-      end = contents.length();
-    }
+    if (content == null) return -1;
 
-    int offset = contents.indexOf(" " + name, start);
+    int offset = content.indexOf(" " + name, start);
     if (offset > -1) {
       return offset + 1;
     }
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
@@ -1,0 +1,63 @@
+package com.sourcegraph.semanticdb_javac;
+
+import com.sun.source.tree.CompilationUnitTree;
+import com.sun.source.tree.LineMap;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.Trees;
+
+import javax.lang.model.element.Element;
+import javax.lang.model.element.Name;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class RangeFinder {
+  public static Optional<Semanticdb.Range> findRange(
+      TreePath path, Trees trees, CompilationUnitTree root, Element element, int startPos) {
+    LineMap lineMap = root.getLineMap();
+    Name name = element.getSimpleName();
+    if (name.contentEquals("<init>")) name = element.getEnclosingElement().getSimpleName();
+
+    int endPos = (int) trees.getSourcePositions().getEndPosition(root, path.getLeaf());
+    // false for anonymous classes
+    if (name.length() != 0) {
+      startPos = findNameIn(root, name, startPos, endPos);
+      endPos = startPos + name.length();
+    }
+
+    if (endPos == -1) {
+      return Optional.empty();
+    }
+
+    Semanticdb.Range range =
+        Semanticdb.Range.newBuilder()
+            .setStartLine((int) lineMap.getLineNumber(startPos) - 1)
+            .setStartCharacter((int) lineMap.getColumnNumber(startPos) - 1)
+            .setEndLine((int) lineMap.getLineNumber(endPos) - 1)
+            .setEndCharacter((int) lineMap.getColumnNumber(endPos) - 1)
+            .build();
+    return Optional.of(range);
+  }
+
+  private static int findNameIn(CompilationUnitTree root, CharSequence name, int start, int end) {
+    String contents;
+    try {
+      contents = root.getSourceFile().getCharContent(true).toString();
+    } catch (IOException e) {
+      return -1;
+    }
+
+    if (end == -1) {
+      end = contents.length();
+    }
+
+    Matcher matcher = Pattern.compile("\\b" + name + "\\b").matcher(contents);
+    matcher.region(start, end);
+    if (matcher.find()) {
+      return matcher.start();
+    } else {
+      return -1;
+    }
+  }
+}
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
@@ -52,12 +52,10 @@ public class RangeFinder {
       end = contents.length();
     }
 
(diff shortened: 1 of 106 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
```diff
@@ -0,0 +1,63 @@
...
+            .setEndLine((int) lineMap.getLineNumber(endPos) - 1)
+            .setEndCharacter((int) lineMap.getColumnNumber(endPos) - 1)
+            .build();
+    return Optional.of(range);
+  }
+
+  private static int findNameIn(CompilationUnitTree root, CharSequence name, int start, int end) {
+    String contents;
+    try {
+      contents = root.getSourceFile().getCharContent(true).toString();
```
olafurpg, line 46: This may be an expensive operation, we should reuse the string between invocations.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (IOException e) {
+      return -1;
+    }
+
+    if (end == -1) {
+      end = contents.length();
+    }
+
+    Matcher matcher = Pattern.compile("\\b" + name + "\\b").matcher(contents);
```
olafurpg, line 55: Regexp can be pretty expensive. It looks like we could get away with `indexOf(" " + name + " ")`

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -103,8 +106,11 @@ public Void visitClass(ClassTree node, Void unused) {
   public Void visitMethod(MethodTree node, Void unused) {
     if (node instanceof JCTree.JCMethodDecl) {
       JCTree.JCMethodDecl meth = (JCTree.JCMethodDecl) node;
-      emitSymbolOccurrence(
-          meth.sym, meth, Role.DEFINITION, CompilerRange.FROM_POINT_TO_SYMBOL_NAME);
+      CompilerRange range = CompilerRange.FROM_POINT_TO_SYMBOL_NAME;
+      if (meth.params.isEmpty() && (meth.sym.flags() & Flags.GENERATEDCONSTR) != 0L) {
```
olafurpg, line 110: Minor: might be worth commenting why this is needed here. It looks like we're specifically targeting only constructors? Is there a difference between this and `name.contentMatches("<init>")`? Is it possible to reuse the logic between this line and the one in `RangeFinder`?

olafurpg (changes requested):
This looks great! Really nice to have this bug fixed. A few minor comments on style and performance.


I recommend running the benchmarks https://sourcegraph.github.io/lsif-java/docs/benchmarks.html just to validate if the optimizations make a difference. I suspect you won't see a difference in caching the file contents from the compilation unit because we compile virtual files in the benchmark, but the speedup might actually be important for build tools that use a different source file implementation.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
Strum355: That would fail to match a generic class of the form `Classname<T>`, but I think just `indexOf(" " + name)` should be good enough

Strum355:
This branch:
```
[info] Benchmark                           (lib)  Mode  Cnt     Score    Error  Units
[info] CompileBench.compile            bytebuddy    ss   10  1371.189 ± 46.755  ms/op
[info] CompileBench.compile                guava    ss   10  1860.100 ± 94.912  ms/op
[info] CompileBench.compileSemanticdb  bytebuddy    ss   10  1684.466 ± 32.033  ms/op
[info] CompileBench.compileSemanticdb      guava    ss   10  2292.896 ± 45.351  ms/op
```

Current main:
```
[info] Benchmark                           (lib)  Mode  Cnt     Score    Error  Units
[info] CompileBench.compile            bytebuddy    ss   10  1379.276 ± 37.184  ms/op
[info] CompileBench.compile                guava    ss   10  1870.692 ± 95.841  ms/op
[info] CompileBench.compileSemanticdb  bytebuddy    ss   10  1662.590 ± 39.720  ms/op
[info] CompileBench.compileSemanticdb      guava    ss   10  2282.368 ± 70.845  ms/op
```

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
Strum355: The difference being, my thinking at the time was that only the no-param constructor would ever be generated, but I realise now this doesn't hold true for the new `record` class types, so I'll revisit this line.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
Strum355: I did a tiny bit of hunting through the jdk types, I saw some references to caching but am not sure if those types are the ones used in our code paths. The benchmarks however appear to show no adverse performance effects

olafurpg (approved):
Awesome, please ignore the perf comments. 

LGTM 👍

olafurpg:
Actually, we should still reuse the string for the compilation unit source code because it's unlikely to appear as a bottleneck in our benchmarks.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
Strum355: Now caching computation of such :+1:

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
olafurpg: We can BTW add Java 15 code to the test suite. Alternatively, we can change the default minimized project to Java 15

olafurpg: I wouldn't worry too much about Java 15 however, we can deal with it later. I suspect very few users are on Java 15 right now.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
+
+public class RangeFinder {
+  private static final HashMap<CompilationUnitTree, String> contentsCache = new HashMap<>();
```
olafurpg, line 19: We should avoid static caches at all costs because they end up holding lots of memory. I don't think we even need a cache, we can store `SemanticdbVisitor.semanticdbText()` in a field.

olafurpg (changes requested):

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
Strum355: Yep, this shouldve been obvious :facepalm: Not sure I follow with `semanticdbText()` though :thinking:

olafurpg: I think we can store the source code string in the constructor here https://github.com/sourcegraph/lsif-java/blob/48bce73664383381c85211a781a427924abae1a2/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java#L48

Strum355: Ah yes I see that now :+1:

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -213,13 +230,15 @@ private String semanticdbSymbol(Symbol sym) {
     }
   }
 
-  private String semanticdbText() {
-    if (!options.includeText) return "";
+  private String semanticdbText(boolean noop) {
```
olafurpg, line 233: I think we can leave this method unchanged.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -39,13 +41,15 @@ public SemanticdbVisitor(
...
     this.options = options;
+    this.trees = Trees.instance(task);
     if (event.getCompilationUnit() instanceof JCTree.JCCompilationUnit) {
       this.endPosTable = ((JCTree.JCCompilationUnit) event.getCompilationUnit()).endPositions;
     } else {
       this.endPosTable = new EmptyEndPosTable();
     }
     this.occurrences = new ArrayList<>();
     this.symbolInfos = new ArrayList<>();
+    // this.source = "";
```
olafurpg, line 52:
```suggestion
    this.source = semanticdbText();
```

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -56,7 +60,7 @@ public SemanticdbVisitor(
         .setSchema(Semanticdb.Schema.SEMANTICDB4)
         .setLanguage(Semanticdb.Language.JAVA)
         .setUri(semanticdbUri())
-        .setText(semanticdbText())
+        .setText(semanticdbText(!options.includeText))
```
olafurpg, line 63:
```suggestion
        .setText(options.includeText ? this.source : "")
```

olafurpg (changes requested):
Minor comments

olafurpg (approved):
LGTM 👍 You may need to run `fixAll` to please the CI.

This is ready to merge 🚀

PR #125, opened 2021-03-16
emit definition occurences for type parameters

Strum355:
This is untested beyond eyeballing the snapshot results. Was done last minute at 10pm before logging off the day before OOO :grin:


olafurpg (approved):
LGTM 🚀

PR #126, opened 2021-03-19
Emit SemanticDB types and signatures

Strum355:
Emit [Type](https://scalameta.org/docs/semanticdb/specification.html#type-2) and [Signature](https://scalameta.org/docs/semanticdb/specification.html#signature-2) messages for symbol occurrences. 
With this information, we can reconstruct accurate hover signatures when converting from SemanticDB to LSIF.

---

**Footnotes**

Wildcard (existential) types proved most difficult to decipher. My work here is a combination of [scalameta Javacp.scala file](https://sourcegraph.com/github.com/scalameta/scalameta/-/blob/semanticdb/metacp/src/main/scala/scala/meta/internal/javacp/Javacp.scala#L453), the semanticdb spec ("Hardlinking comes in handy in situations when an advanced type such as ... `ExistentialType`") and this one table entry from the semanticdb spec (in which the params are the opposite way around): 
image

As wildcard params dont have a symbol emitting for them, I create a hardlink as opposed to a symlink in the declarations field of the existential type. In a mixed wildcard and non-wildcard scenario, the declarations ordering must be derived from the `ExistentialType.tpe.type_arguments` field, as in the case of `Triplet<?, String, ? extends Integer>`, the ordering of declarations would be `symlink: "String", hardlinks: "local_wildcard", hardlinks: "local_wildcard"+upper_bound: "java/lang/Integer#"`. The relative ordering of symlinks to symlinks and hardlinks to hardlinks is preserved, just not symlinks to hardlinks and vice versa, so there should be no problem there. 

I had to do some guessing as to the behaviour when theres multiple wildcard type params for a given type (as is the case with e.g. `Map<?, ?>`), which isn't covered in the semanticdb spec, but this seems fairly sane.

<details>
<summary>Decoded Protobuf</summary>

Link to Java source for below: https://github.com/sourcegraph/lsif-java/pull/126/files#diff-3f725eecc09c1818a18120a2d471018890dd664200b233dcbee8c14c650824aeR10

```clojure
  symbols {
    symbol: "minimized/ParameterizedTypes#doStuff()."
    signature {
      method_signature {
        type_parameters {
        }
        parameter_lists {
        }
        return_type {
          existential_type {
            tpe {
              type_ref {
                symbol: "java/util/Map#"
                type_arguments {
                  type_ref {
                    symbol: "local_wildcard"
                  }
                }
                type_arguments {
                  type_ref {
                    symbol: "local_wildcard"
                  }
                }
              }
            }
            declarations {
              hardlinks {
                symbol: "local_wildcard"
                signature {
                  type_signature {
                    type_parameters {
                    }
                    upper_bound {
                      type_ref {
                        symbol: "java/lang/String#"
                      }
                    }
                  }
                }
              }
              hardlinks {
                symbol: "local_wildcard"
                signature {
                  type_signature {
                    type_parameters {
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
```

</details>

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
@@ -11,11 +11,14 @@ import javax.lang.model.util.SimpleTypeVisitor8;
 import java.util.ArrayList;
 import java.util.List;
 
-import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
-
 public final class SemanticdbSignatures {
   private final GlobalSymbolsCache cache;
   private final LocalSymbolsCache locals;
+
+  private static final Semanticdb.Type UNRESOLVED_TYPE_REF =
+      Semanticdb.Type.newBuilder()
+          .setTypeRef(TypeRef.newBuilder().setSymbol("unresolved_type#"))
+          .build();
 
   public SemanticdbSignatures(GlobalSymbolsCache cache, LocalSymbolsCache locals) {
     this.cache = cache;
@@ -43,17 +46,14 @@ public final class SemanticdbSignatures {
     if (sym.getSuperclass() != Type.noType) {
       Semanticdb.Type superType = generateType(sym.getSuperclass());
       if (superType == null) {
-        pprint("CLASS GAVE NULL " + sym.getSuperclass() + " " + sym);
-      } else {
-        builder.addParents(superType);
+        superType = UNRESOLVED_TYPE_REF;
       }
+      builder.addParents(superType);
     }
     for (Type iType : sym.getInterfaces()) {
       Semanticdb.Type type = generateType(iType);
-      // this can happen if an interface type isnt properly resolved. Shows as _root_ in snapshots
       if (type == null) {
-        pprint("INTERFACE GAVE NULL " + iType + " " + sym);
-        continue;
+        type = UNRESOLVED_TYPE_REF;
       }
       builder.addParents(type);
     }
@@ -85,8 +85,7 @@ public final class SemanticdbSignatures {
   private Signature generateFieldSignature(Symbol.VarSymbol sym) {
     Semanticdb.Type generateType = generateType(sym.type);
     if (generateType == null) {
-      pprint("FIELD TYPE GAVE NULL " + sym.type + " " + cache.semanticdbSymbol(sym, locals));
-      return null;
+      generateType = UNRESOLVED_TYPE_REF;
     }
     return Signature.newBuilder()
         .setValueSignature(ValueSignature.newBuilder().setTpe(generateType))
@@ -100,7 +99,7 @@ public final class SemanticdbSignatures {
 
     Semanticdb.Type upperBound = generateType(sym.type.getUpperBound());
     if (upperBound != null) builder.setUpperBound(upperBound);
-    if (upperBound == null) pprint("UPPER BOUND GAVE NULL " + sym + " " + sym.type.getUpperBound());
+    else builder.setUpperBound(UNRESOLVED_TYPE_REF);
 
     return Signature.newBuilder().setTypeSignature(builder).build();
   }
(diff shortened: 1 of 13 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
```diff
@@ -0,0 +1,265 @@
...
+                    .setDeclarations(declarations))
+            .build();
+      }
+    }
+
+    @Override
+    public Semanticdb.Type visitArray(ArrayType t, Void unused) {
+      Semanticdb.Type arrayComponentType = super.visit(t.getComponentType());
+      if (arrayComponentType == null) {
+        pprint("ARRAY GAVE NULL " + t.getComponentType() + " " + t);
```
olafurpg, line 194: Remove?

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -54,6 +56,7 @@ public SemanticdbVisitor(
   }
 
   public Semanticdb.TextDocument buildTextDocument(CompilationUnitTree tree) {
+    pprint(semanticdbUri());
```
olafurpg, line 59: Remove?

olafurpg (commented):
This is exciting progress! The implementation looks good to me. Just missing the pretty-printer before we can merge this. It's best to wait until the lsif-semanticdb PR is merged before we integrate the pretty-printer into LSIF hover messages

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
Strum355: Im wondering if, for cases like this, we should add some magic symbol as with `"local_wildcard"`, but instead `"unresolved"`. I would personally prefer this in an editor e.g. `class Type extends [unresolved]` as opposed to leaving me to believe that it extends nothing and is a correct hover of `class Type`. What do you think?

olafurpg: I think it's OK to return a magic symbol instead of silently failing. It would make it easier to diagnose a bug/problem if a user encounters it. I think it's also OK to return null and skip the signature in these cases.

Strum355: https://github.com/sourcegraph/lsif-java/pull/126/commits/dcb3cbbd06245934426cfcd4cff16ddea9b45123 Let me know what you think of this change

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
```diff
@@ -11,12 +11,15 @@
...
 
-import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
-
 public final class SemanticdbSignatures {
   private final GlobalSymbolsCache cache;
   private final LocalSymbolsCache locals;
 
+  private static final Semanticdb.Type UnresolvedTypeRef =
+      Semanticdb.Type.newBuilder()
+          .setTypeRef(TypeRef.newBuilder().setSymbol("[unresolved]"))
```
olafurpg, line 20: Let's use a valid SemanticDB symbol like `local_unresolved` instead

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbSignatures.java
```diff
@@ -11,12 +11,15 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import static com.sourcegraph.semanticdb_javac.Debugging.pprint;
-
 public final class SemanticdbSignatures {
   private final GlobalSymbolsCache cache;
   private final LocalSymbolsCache locals;
 
+  private static final Semanticdb.Type UnresolvedTypeRef =
```
olafurpg, line 18: I believe the standard naming convention for static variables is s/UnresolvedTypeRef/UNRESOLVED_TYPE_REF/

(GitHub is not allowing me to write suggestions for some reason)

PR #127, opened 2021-03-24
Implement LSIF emitter in Java.

olafurpg:
Previously, we shelled out to `lsif-semanticdb` to convert SemanticDB
files into LSIF. This commit ports the original Go implemementation to
Java so that we can call a normal method instead of shelling out to a
separate process.

The biggest problem with the previous solution is that it complicated
installation for users. As we evolve lsif-java, the conversion logic
will also evolve and we want to avoid the situation where users have
an outdated version of lsif-semanticdb installed. With this commit,
users only need to install `lsif-java` and everything should work
correctly.

This PR also adds a new `snapshot-lsif` command to pretty-print an
LSIF index, similar to how we pretty-print SemanticDB payloads. This
command is mostly useful for testing purposes, but could be helpful
to troubleshoot user issues.

LSIF snapshot generated for lsif-go

CleanShot 2021-03-24 at 11 48 33@2x

LSIF snapshot generated for lsif-node

CleanShot 2021-03-24 at 11 48 53@2x


lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
```diff
@@ -17,7 +17,7 @@ object SemanticdbPrinters {
       .linesWithSeparators
       .zipWithIndex
       .foreach { case (line, i) =>
-        out.append(line)
+        out.append(line.replace("\t", "→"))
```
Strum355, line 20: Im curious if theres a reason we cant do `replace("\t", "    ")` instead here.

olafurpg: I we can do that but it removes the fact that there's a tab character, which may be helpful to debug an issue.

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
```diff
@@ -0,0 +1,132 @@
...
+      rangeIds.add(rangeId);
+
+      // Range
+      writer.emitNext(rangeId, ids.resultSet);
+
+      // Reference
+      writer.emitItem(ids.referenceResult, rangeId, doc.id);
+
+      // Definition
+      if (occ.getRole() == SymbolOccurrence.Role.DEFINITION && ids.isDefinitionDefined()) {
```
Strum355, line 104: in what case is `ids.isDefinitionDefined() == true` aka when would we have two identical definitions?

Strum355 (approved):

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
olafurpg: The definition ID should always be available in this case, it's a bug if it's not. I updated the code to report an error message in this case instead of silently ignoring it.

PR #131, opened 2021-03-29
Add signature pretty-printer to lsif-semanticdb conversion process

Strum355:
PR adds support for generating a pretty-printed hover text from the emitted SemanticDB Signature and SymbolInformation types.
This happens at the SemanticDB to LSIF stage and is included in the hover results lists for a range, alongside the doc string if it exists.

---

TODO (either in this PR or a follow-up):
- ~~annotation interfaces~~
- annotations on SymbolInformation (**follow up**)
- additional properties (strictfp/synchronized etc) (**follow up**)
- ~~pretty-print enums properly~~
- ~~implements vs extends~~
- Comment outlined [here](https://github.com/sourcegraph/lsif-java/pull/131#discussion_r605485119) (**follow up**)
- ~~skip `abstract` in interface + interface methods~~
- ~~constructors~~
- ~~`this` keyword references field named `this` that doesn't exist~~


lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
```diff
@@ -37,7 +43,9 @@ object SemanticdbPrinters {
   private def formatOccurrence(
       out: StringBuilder,
       occ: SymbolOccurrence,
-      line: String
+      line: String,
+      symbols: Map[String, SymbolInformation],
```
olafurpg, line 47: What is the difference between this map and `Symtab`? Would be nice to have only one symtab

lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
```diff
@@ -76,6 +84,19 @@ object SemanticdbPrinters {
...
           ""
       )
+      .append(
+        symbols.get(occ.getSymbol) match {
+          case Some(info) if occ.getRole == Role.DEFINITION =>
+            val sig = new SignatureFormatter(info, symtab).formatSymbol()
+            if (sig == "")
+              " " + info.getDisplayName
+            else
+              " " + sig + ""
```
olafurpg, line 94:
```suggestion
              " " + sig
```

lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
```diff
@@ -76,6 +84,19 @@ object SemanticdbPrinters {
         else
           ""
       )
+      .append(
+        symbols.get(occ.getSymbol) match {
+          case Some(info) if occ.getRole == Role.DEFINITION =>
+            val sig = new SignatureFormatter(info, symtab).formatSymbol()
+            if (sig == "")
```
olafurpg, line 91: nit:
```suggestion
            if (sig.isEmpty)
```

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -119,6 +120,31 @@ object SnapshotLsifCommand {
...
+                hover <- lsif.hoverVertexes.get(hoverId)
+              } yield hover
+            ).get
+              .getContentsList
+              .asScala
+              .find { h =>
+                h.getLanguage != Language.UNKNOWN_LANGUAGE.toString.toLowerCase
+              }
+              .get
+              .getValue
```
olafurpg, line 139: I think this refactoring makes it a bit easier to read and it handles the case when you have multiple hovers

```suggestion
          val hover =
            (
              for {
                resultSetId <- lsif.next.get(o.getId).toList
                hoverId <- lsif.hoverEdges.get(resultSetId).toList
                hover <- lsif.hoverVertexes.get(hoverId).toList
                contents <- hover.getContentsList().asScala
                if contents.getLanguage !=
                  Language.UNKNOWN_LANGUAGE.toString.toLowerCase
              } yield contents.getValue()
            ).mkString("\n")
```

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -119,6 +120,31 @@ object SnapshotLsifCommand {
...
+          val symInfo = SymbolInformation
+            .newBuilder()
+            // we cheese it a bit here, as this is less work than trying to reconstruct
+            // a Signature from the pretty-printed Signature, with accompanying logic
+            // to fallback to display_name in SemanticdbPrinters.scala
+            .setDisplayName(hover).setSymbol(symbol).build()
```
olafurpg, line 145: Can we use `Documentation` instead?

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
```diff
@@ -112,9 +114,21 @@ private Integer processDocument(LsifTextDocument doc, Set<String> isExportedSymb
         }
 
         // Hover
+        ArrayList<MarkedString> markedStrings = new ArrayList<>();
         String documentation = symbolInformation.getDocumentation().getMessage();
         if (!documentation.isEmpty()) {
-          int hoverId = writer.emitHoverResult(doc.semanticdb.getLanguage(), documentation);
+          markedStrings.add(new MarkedString(Semanticdb.Language.UNKNOWN_LANGUAGE, documentation));
```
olafurpg, line 120: Why do we use unknown language here?

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifWriter.java
```diff
@@ -90,15 +93,21 @@ public int emitDefinitionResult(int resultSet) {
     return definitionResult;
   }
 
-  public int emitHoverResult(Semanticdb.Language language, String value) {
+  public int emitHoverResult(MarkedString... markedStrings) {
```
olafurpg, line 96: It doesn't look like we use the varargs syntax when calling this method so this should be safe to change
```suggestion
  public int emitHoverResult(MarkedString[] markedStrings) {
```

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java
```diff
@@ -0,0 +1,309 @@
...
+    switch (symbolInformation.getKind()) {
+      case CLASS:
+        if (has(Property.ENUM)) {
+          printKeyword("enum");
+        } else {
+          printKeyword("class");
+        }
+        break;
+      case INTERFACE:
+        if (isAnnotation) {
```
olafurpg, line 64: I think it would make sense to have a property or kind for annotation interfaces in SemanticDB. It's good to know this workaround and it's fine to use it for now

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java
```diff
@@ -0,0 +1,309 @@
...
+
+    List<SymbolInformation> typeParameters = getSymlinks(classSignature.getTypeParameters());
+    if (!typeParameters.isEmpty()) {
+      s.append(
+          typeParameters.stream()
+              .map(this::formatTypeParameter)
+              .collect(Collectors.joining(", ", "<", ">")));
+    }
+
+    List<Type> parentsList =
```
olafurpg, line 81: nit: the name "parentsList" would hint this is just normal parents 
```suggestion
    List<Type> nonSyntheticParents =
```

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    printKeyword(formatType(methodSignature.getReturnType()));
+    s.append(symbolInformation.getDisplayName());
+
+    s.append(
+        methodSignature.getParameterListsList().stream()
+            .flatMap((params) -> getSymlinks(params).stream())
+            .map(
+                (symInfo) ->
```
olafurpg, line 123: FYI, I don't think you need to wrap the parameter in parentheses
```suggestion
                symInfo ->
```

tests/snapshots/src/main/generated/com/airbnb/epoxy/AsyncEpoxyController.java
```diff
@@ -40,8 +40,8 @@ public AsyncEpoxyController() {
    *                    both on the main thread.
    */
   public AsyncEpoxyController(boolean enableAsync) {
-//       ^^^^^^ definition com/airbnb/epoxy/AsyncEpoxyController#`<init>`(+1).
-//                                    ^^^^^^^^^^^ definition local0
+//       ^^^^^^ definition com/airbnb/epoxy/AsyncEpoxyController#`<init>`(+1). public <init>(boolean enableAsync)
```
olafurpg: What do you think about showing the class name in this case?
```suggestion
//       ^^^^^^ definition com/airbnb/epoxy/AsyncEpoxyController#`<init>`(+1). public AsyncEpoxyController(boolean enableAsync)
```

tests/snapshots/src/main/generated/com/airbnb/epoxy/DiffResult.java
```diff
@@ -64,16 +64,16 @@
...
   /** No changes were made to the models. */
   static DiffResult noOp(@Nullable List<? extends EpoxyModel<?>> models) {
 //       ^^^^^^^^^^ reference com/airbnb/epoxy/DiffResult#
-//                  ^^^^ definition com/airbnb/epoxy/DiffResult#noOp().
+//                  ^^^^ definition com/airbnb/epoxy/DiffResult#noOp(). static DiffResult noOp(List<? extends EpoxyModel<?>> models)
 //                        ^^^^^^^^ reference androidx/annotation/Nullable#
 //                                 ^^^^ reference java/util/List#
 //                                                ^^^^^^^^^^ reference com/airbnb/epoxy/EpoxyModel#
-//                                                               ^^^^^^ definition local0
+//                                                               ^^^^^^ definition local0 List<? extends EpoxyModel<?>> models
```
olafurpg, line 76: 👏 🤩

tests/snapshots/src/main/generated/com/airbnb/epoxy/EpoxyAsyncUtil.java
```diff
@@ -35,17 +35,17 @@
...
+//        ^^^^^^ definition com/airbnb/epoxy/EpoxyAsyncUtil#`<init>`(). private <init>()
   }
 
   /**
    * A Handler class that uses the main thread's Looper.
    */
   public static final Handler MAIN_THREAD_HANDLER =
 //                    ^^^^^^^ reference _root_/
-//                            ^^^^^^^^^^^^^^^^^^^ definition com/airbnb/epoxy/EpoxyAsyncUtil#MAIN_THREAD_HANDLER.
+//                            ^^^^^^^^^^^^^^^^^^^ definition com/airbnb/epoxy/EpoxyAsyncUtil#MAIN_THREAD_HANDLER. public static final unresolved_type MAIN_THREAD_HANDLER
```
olafurpg, line 48: Why do we have `unresolved_type` here?

tests/snapshots/src/main/generated/com/airbnb/epoxy/AsyncEpoxyController.java
olafurpg: We can get the symbol owner's descriptor using the new symbol parser

olafurpg (changes requested):
This is outstanding work! It will be a big milestone when lsif-java supports proper hover messages.

A few comments, mostly minor. The only blocking comments are

- print class name instead of `<init>`
- why are there so many `unresolved_type` in the snapshots?

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
Strum355: Im a bit unsure about how to approach this one. 

- On one side, by setting it to unknown language, we can extract the proper pretty printed signature for the SnapshotLsifCommand (explained [here](https://github.com/sourcegraph/lsif-java/pull/131/files/d82dd45e3c3f20d429b4c7b1f47d66824b7aa449#diff-94acd677e36ffbda144a22796bd114db1af4154de5513b67a5241a83a02df453R142)
- On the other hand, we may lose syntax highlighting of the javadoc comment. This one's a bit of a double edged sword, as depending on the highlighter, it may end up highlighting words that are not meant to be (like `this` when used in a sentence as opposed to a code snippet in the javadoc comment)

Strum355:
> why are there so many unresolved_type in the snapshots?

There seems to be some issue with types from the android sdk occasionally (and Im not sure why only occasionally) where they dont resolved. This is always accompanied with the reference occurrence being given the symbol `_root_/`. Id be curious to see if this happens in non-android-based codebases too.

olafurpg (approved):
LGTM 👍 Feel free to merge once you think this is ready to go.

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
Strum355: Its weird because many other android types _do_ resolve. :shrug:

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
Strum355: That sounds good for a follow-up PR after discussing in our 1:1 :+1:

PR #132, opened 2021-03-31
Update documentation

olafurpg:
- remove mention of the native image, which we no longer upload
- reformulate the installation docs

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -11,41 +11,22 @@ tools that we're planning to support in the future.
 
 ## Install `lsif-java`
 
-Most users only install `lsif-java` on a CI machine to upload LSIF indexes to a
-remote service like [Sourcegraph](https://sourcegraph.com/). The easiest way to
-install `lsif-java` is to download the native binary. However, you can also
-install `lsif-java` as a Java binary if that's easier to integrate with your
-setup.
-
-### Native binary
-
-The native binary is only available for Linux and macOS. The native binary
-includes all dependencies and does not need further access to the internet after
-it's been downloaded.
-
-```sh
-# macOS
-curl -Lo lsif-java https://github.com/sourcegraph/lsif-java/releases/download/v@STABLE_VERSION@/lsif-java-x86_64-apple-darwin \
-  && chmod +x lsif-java \
-  && ./lsif-java --help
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -55,17 +36,38 @@ curl -fLo coursier https://git.io/coursier-cli \
...
-which includes all dependencies and does not require further access to the
-internet after installation.
+Use the `lsif-java` fat jar if your CI does not allow downloading binaries from
+the internet.
+
+Use the [Coursier](https://get-coursier.io/docs/cli-installation) `bootstrap`
+command to generate a local fat jar binary. The fat jar binary includes all
+dependencies and does not require further access to the internet after
+installation. The local fat jar will somehow need to be made available to your
+CI machine.
```
Strum355, line 58: imo we should supply a fat-jar as a downloadable artifact *as well*.

I would imagine that "if your CI does not allow downloading binaries from the internet" holds true (for security reasons), some companies (especially in certain industries) would have an auditing process involved in getting coursier approved first on top of lsif-java itself, resulting in customers who would prefer to just have a downloadable fat-jar to reduce the entry barrier.

Just my 2c after much back-n-forth with some guy online working at a financial institute looking to adopt Golang and the many hoops they have to go through to add anything new to the developer workflow.

Strum355 (approved):

docs/getting-started.md
olafurpg: The instructions to generate the fat-jar are copy-pasteable commands that you can run anywhere to generate the standalone executable. My issue with uploading fat jars to GitHub releases is that it doesn't provide a clean way for users to upgrade between versions. With the coursier bootstrap command the lsif-java version number is part of the copy-pasted command.

olafurpg: I can imagine that some customers will still ask to have a downloadable fatjar from GitHub releases and maybe we'll just comply because it's easier. It's not complicated to generate a fatjar in CI and upload it

Strum355: Totally, we can strongly recommend coursier for that reason. As you said, releasing fatjars is trivial enough, especially given we're already using github actions :slightly_smiling_face:

PR #151, opened 2021-04-14
fix range building for all constructors

Strum355:
Previously, only generated constructors were delegated to using text search to find the range encompassing the constructor name (aka the class name, which shares the same range as the generated default constructor).
For this reason, non-generated constructors were not hitting the code-path that checks if the `name == "<init>"`, aka it was a constructor, which followed with logic to build a range based on the owning class's name, causing the range to be built based on the synthetic `<init>` name.

This PR fixes this bug by delegating _all_ constructors to the aforementioned text search codepath. It does it slightly differently by using text search from point, as constructors' preferred/point position starts from the name, avoiding the problem outlined [here](https://github.com/sourcegraph/lsif-java/pull/151#discussion_r613316106).

This allows for a minor simplification in the `RangeFinder` class, as we no longer have to prepend the search name with a space. This was only required for constructors which may have the constructor name at the start of the line (after whitespace).

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
@@ -30,6 +30,9 @@ public class RangeFinder {
     int endPos = (int) trees.getSourcePositions().getEndPosition(root, path.getLeaf());
     // false for anonymous classes
     if (name.length() != 0) {
+      // for declared constructors with no modifiers, findNameIn will return -1 as the name will
+      // not start with a space, the first char is the start of the name
+      if (source.charAt(startPos) != ' ') startPos -= 1;
       startPos = findNameIn(name, startPos, source);
       endPos = startPos + name.length();
     }
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
@@ -30,9 +30,6 @@ public class RangeFinder {
     int endPos = (int) trees.getSourcePositions().getEndPosition(root, path.getLeaf());
     // false for anonymous classes
     if (name.length() != 0) {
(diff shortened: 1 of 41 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
```diff
@@ -30,6 +30,9 @@
     int endPos = (int) trees.getSourcePositions().getEndPosition(root, path.getLeaf());
     // false for anonymous classes
     if (name.length() != 0) {
+      // for declared constructors with no modifiers, findNameIn will return -1 as the name will
+      // not start with a space, the first char is the start of the name
+      if (source.charAt(startPos) != ' ') startPos -= 1;
```
olafurpg, line 35: How do we know `startPos -= 1` fixes this problem?

olafurpg (approved):
LGTM 👍 One non-blocking comment about the fix, feel free to merge.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/RangeFinder.java
Strum355: Youre right, that is an assumption, with holes. Its currently based on the assumption that
1. the file uses spaces for indentation
2. the constructor is indented aka is preceded by spaces
3. any type parameter bounds do not have the constructor name as a prefix aka `<T extends Stringer> String()`

An example of where point 3 falls apart can be seen here:
image

Strum355: All is well, constructor preferred position starts at the name of the constructor : ) :tada: 

image

olafurpg (commented):
The new fix looks a lot cleaner, thank you for following up on that comment 🙏

olafurpg:
I'm not sure what's causing the CI failure. It looks like Gradle is failing sporadically 🤔 Might be worth trying to restart the jobs tomorrow.

Strum355:
`sbt test` passes locally for me, still failing in CI :thinking: 
Ill merge this PR and we can investigate CI at a later stage

Issue #154, opened 2021-04-15
Reference occurrences range over the entire constructor call


olafurpg:
https://github.com/sourcegraph/lsif-java/blob/687fef86c382ddedd9aaa462df0ef33589a8dded/tests/snapshots/src/main/generated/com/airbnb/epoxy/AsyncEpoxyDiffer.java#L96-L101

```diff
- obtained
+ expected
      this.executor = new HandlerExecutor(handler);
  //  ^^^^ reference com/airbnb/epoxy/AsyncEpoxyDiffer#
  //       ^^^^^^^^ reference com/airbnb/epoxy/AsyncEpoxyDiffer#executor.
- //                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#`<init>`().
+ //                      ^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#`<init>`().
  //                      ^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#
  //                                      ^^^^^^^ reference local1
```

olafurpg:
As discussed offline, we should also stop generating the occurrence to the type symbol for consistency with VS Code and IntelliJ
```diff
- obtained
+ expected
      this.executor = new HandlerExecutor(handler);
  //  ^^^^ reference com/airbnb/epoxy/AsyncEpoxyDiffer#
  //       ^^^^^^^^ reference com/airbnb/epoxy/AsyncEpoxyDiffer#executor.
- //                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#`<init>`().
+ //                      ^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#`<init>`().
-  //                      ^^^^^^^^^^^^^^^ reference com/airbnb/epoxy/HandlerExecutor#
  //                                      ^^^^^^^ reference local1
```

PR #156, opened 2021-04-15
fix ctor range via textsearch & dont emit ref to class in ctro call

Strum355:
Previously, constructor calls would also emit a reference range for the identifier, pointing to the class. Additionally, the range for a constructor call included all the (type) parameters too.

This PR addresses this this by skipping scanning the identifier and by using text-search range finding.

Closes https://github.com/sourcegraph/lsif-java/issues/154

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -221,9 +221,14 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
   public Void visitNewClass(NewClassTree node, Void unused) {
     if (node instanceof JCTree.JCNewClass) {
       JCTree.JCNewClass cls = (JCTree.JCNewClass) node;
-      emitSymbolOccurrence(cls.constructor, cls, Role.REFERENCE, CompilerRange.FROM_START_TO_END);
+      emitSymbolOccurrence(cls.constructor, cls, Role.REFERENCE, CompilerRange.FROM_TEXT_SEARCH);
     }
-    return super.visitNewClass(node, unused);
+
+    // to avoid emitting a reference to the class itself, we manually scan everything
+    // except the identifier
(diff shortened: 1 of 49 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -221,9 +221,14 @@ public Void visitMemberSelect(MemberSelectTree node, Void unused) {
   public Void visitNewClass(NewClassTree node, Void unused) {
     if (node instanceof JCTree.JCNewClass) {
       JCTree.JCNewClass cls = (JCTree.JCNewClass) node;
-      emitSymbolOccurrence(cls.constructor, cls, Role.REFERENCE, CompilerRange.FROM_START_TO_END);
+      emitSymbolOccurrence(cls.constructor, cls, Role.REFERENCE, CompilerRange.FROM_TEXT_SEARCH);
```
olafurpg, line 224: Can we actually pass in the identifier to avoid text search?
```suggestion
      emitSymbolOccurrence(cls.constructor, cls.getIdentifier(), Role.REFERENCE, CompilerRange.FROM_START_TO_END);
```

olafurpg (approved):
LGTM 👍 That was a quick fix, one minor comment. Feel free to merge

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
Strum355: That seems to also include the type parameters interestingly :thinking: 
image

Strum355: And potentially erroneous results :grimacing: 

image

olafurpg: Interesting, text search seems to be the way to go then.

PR #158, opened 2021-04-15
Reduce emitted occurences noise from anonymous classes

Strum355:
This PR addresses the noisy output to synthetic elements arising from anonymous classes and anonymous class constructors.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,6 +104,7 @@ lazy val plugin = project
     javaOnlySettings,
     moduleName := "semanticdb-javac",
     javaToolchainVersion := "8",
+    javacOptions += "-g",
     assemblyShadeRules.in(assembly) :=
       Seq(
         ShadeRule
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -103,6 +103,7 @@ lazy val plugin = project
     javaOnlySettings,
     moduleName := "semanticdb-javac",
     javaToolchainVersion := "1.8",
+    javacOptions += "-g",
```
olafurpg, line 106: Is this required for the fix or it it only needed for debugging?

tests/snapshots/src/main/generated/com/airbnb/epoxy/EpoxyTouchHelper.java
```diff
@@ -403,26 +403,18 @@ public ItemTouchHelper andCallbacks(final DragCallbacks<U> callbacks) {
 //                    ^^^^^^^^^^^^^^^ definition local18 unresolved_type itemTouchHelper
           new ItemTouchHelper(new EpoxyModelTouchCallback<U>(controller, targetModelClass) {
```
olafurpg, line 404: Nice!

olafurpg (approved):
Awesome, really good to see this fixed 👍

build.sbt
Strum355: Not required, but makes intellij not show a constant warning in the variables view while debugging. 
Unless you've a reason we shouldn't include it, I dont see it as anything but benign

olafurpg: Sounds good, was just curious if this flag affected semanticdb

PR #160, opened 2021-04-15
Pretty-print `throws` clause in method signatures

### LINKED ISSUE #147: Add `throws` clause to pretty-printed signature
e.g. `public void method() throws OutOfBoundsException` currently does not include the `throws` in the signature

Strum355:
Add 'throws' field to `MethodSignature` in semanticdb for implementing pretty-printing, to be upstreamed to semanticdb spec. 
This is a fairly trivial change so it shouldnt be a big deal to upstream.

Closes #147

### DIFF
--- a/tests/minimized/src/main/java/minimized/Methods.java
+++ b/tests/minimized/src/main/java/minimized/Methods.java
@@ -17,7 +17,7 @@ public class Methods {
     return value + "1";
   }
 
-  public static String app(int n, String m) {
+  public static String app(int n, String m) throws RuntimeException, IndexOutOfBoundsException {
     Methods methods = new Methods();
     int a = staticOverload(n);
     String b = staticOverload(m);
--- a/lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java
+++ b/lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/SignatureFormatter.java
@@ -178,6 +178,14 @@ public class SignatureFormatter {
                         + " "
                         + symInfo.getDisplayName())
             .collect(Collectors.joining(", ", "(", ")")));
+
+    if (!methodSignature.getThrowsList().isEmpty()) {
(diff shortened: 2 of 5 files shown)


tests/minimized/src/main/java/minimized/Methods.java
```diff
@@ -17,7 +17,7 @@ private static String staticOverload(String value) {
     return value + "1";
   }
 
-  public static String app(int n, String m) {
+  public static String app(int n, String m) throws RuntimeException {
```
olafurpg, line 20: I'm surprised that epoxy doesn't use `throws` anywhere. Checked exceptions are IMO one of the biggest pains of working with Java. Can we add one test case where the exception has type parameters just to stress the code path that `throws` is `Type` and not symbols

olafurpg (changes requested):
I'm investigating in #163 why the CI is failing. 

One minor comments, otherwise looks good

tests/minimized/src/main/java/minimized/Methods.java
Strum355: Turns out custom exceptions (that extend Exception/Throwable) cant have type parameters :thinking:

olafurpg: Where did you find examples with type parameters? Are they only in the JDK?

Strum355: https://sourcegraph.com/search?q=context:global+%28+throws+%5Ba-zA-Z%5D%2B%3C%5BA-Z%5D%5Ba-zA-Z%5D*%3E%29+lang:java&patternType=regexp&case=yes

Strum355: Not exactly "conclusive" or "proving" evidence by any means :grin:

Strum355: Ive added an additional exception to the signature to make up for it. Gonna merge :+1:

PR #171, opened 2021-04-19
fix range building for tabs or mixed tabs/spaces files

### LINKED ISSUE #115: Wrong ranges for lines with tab indentation
In the given file [here](https://sourcegraph.com/github.com/CS3305-Team-2/ReFund-Backend@da112293426515de208464aefdbeb3c845d725e2/-/blob/src/main/java/com/bestteam/models/Awards.java), certain normal methods only return [search based results](https://sourcegraph.com/github.com/CS3305-Team-2/ReFund-Backend@da112293426515de208464aefdbeb3c845d725e2/-/blob/src/main/java/com/bestteam/models/Awards.java#L38:14) while others give us [semantic based results](https://sourcegraph.com/github.com/CS3305-Team-2/ReFund-Backend@da112293426515de208464aefdbeb3c845d725e2/-/blob/src/main/java/com/bestteam/models/Awards.java#L66:19) (albeit with "no hover info").

These are all completely basic methods taking and returning basic types (Long, String etc). 

Fringe thought, is this related to `@Entity` or `@Table` annotations? There doesn't appear to be a pattern between getters vs setters however.

### LINKED ISSUE #169: Handle TABs in code
The position info for code that uses TABs :vomiting_face: seems to be off.

Example source file...
```java
public class BarJava
{
 private Integer i;
→private Integer j;
}
```
To be clear...
`i` variable definition has a single space before `private`
`j` variable definition has a single tab before `private`

Resulting snapshot index...
```
public class BarJava
//           ^^^^^^^ definition javafoo/BarJava# public class BarJava
//           ^^^^^^^ definition javafoo/BarJava#`<init>`(). public BarJava()
{
 private Integer i;
//       ^^^^^^^ reference java/lang/Integer#
//               ^ definition javafoo/BarJava#i. private Integer i
→private Integer j;
//              ^^^^^^^ reference java/lang/Integer#
//                      ^ definition javafoo/BarJava#j. private Integer j
}
```

`i` position looks fine
`j` position looks out

Strum355:
Previously, ranges would be wrong for files indented with tabs or tabs&spaces.

This happened because javac replaces every tab with 8 spaces in the linemap. As this is potentially inconsistent with the source file itself, we adjust for that when building ranges if the line is actually indented with tabs.

As for every tab there are 8 spaces, we remove 7 spaces for every tab to get the correct char offset (note: different to _column_ offset your editor shows)

image

Closes #115 and closes #169

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -230,16 +230,19 @@ lazy val minimized = project
   .in(file("tests/minimized/.j11"))
   .settings(minimizedSettings)
   .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)
 
 lazy val minimized8 = project
   .in(file("tests/minimized/.j8"))
   .settings(minimizedSettings, javaToolchainVersion := "8")
   .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)
 
 lazy val minimized15 = project
   .in(file("tests/minimized/.j15"))
   .settings(minimizedSettings, javaToolchainVersion := "15")
   .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)
 
 lazy val minimizedScala = project
   .in(file("tests/minimized-scala"))
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -255,6 +255,7 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
 
   private Optional<Semanticdb.Range> semanticdbRange(
       JCDiagnostic.DiagnosticPosition pos, CompilerRange kind, Symbol sym) {
+    LineMap lineMap = event.getCompilationUnit().getLineMap();
     if (sym == null) return Optional.empty();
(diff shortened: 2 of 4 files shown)


olafurpg (changes requested):
The fix looks good. Can we add some test case to reproduce the issue?

We may need to disable javafmt in the `minimized` project so that we can write tab characters in the source files.
```diff
lazy val minimized = project
  .in(file("tests/minimized/.j11"))
  .settings(minimizedSettings)
  .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)

lazy val minimized8 = project
  .in(file("tests/minimized/.j8"))
  .settings(minimizedSettings, javaToolchainVersion := "8")
  .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)

lazy val minimized15 = project
  .in(file("tests/minimized/.j15"))
  .settings(minimizedSettings, javaToolchainVersion := "15")
  .dependsOn(agent, plugin)
+  .disablePlugins(JavaFormatterPlugin)
```

Strum355:
@olafurpg I've added a snapshot example just now. Thankfully as we're emitting char offset ranges as opposed to column offset ranges, it didnt require any changes to the snapshot printer to make everything line up just right :slightly_smiling_face:

olafurpg (approved):
LGTM 🔥

PR #172, opened 2021-04-19
emit overridden methods for method SymbolInformation

Strum355:
Previously, if `A` is a subclass of/implements `B` and `A` overrides `B.X`, find references on `A.X` would not include the references for the `B.X` and vice versa.

Addresses https://github.com/sourcegraph/sourcegraph/issues/15699 for lsif-java.

This PR is pending validation of https://github.com/sourcegraph/sourcegraph/pull/20160 with lsif-java output specifically.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -5,6 +5,7 @@ import com.sun.source.util.*;
 import com.sun.tools.javac.code.Flags;
 import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Type;
+import com.sun.tools.javac.model.JavacTypes;
 import com.sun.tools.javac.tree.EndPosTable;
 import com.sun.tools.javac.tree.JCTree;
 import com.sun.tools.javac.util.JCDiagnostic;
@@ -19,10 +20,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import static com.sourcegraph.semanticdb_javac.SemanticdbTypeVisitor.ARRAY_SYMBOL;
@@ -34,6 +32,7 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
   private final LocalSymbolsCache locals;
   private final JavacTask task;
   private final TaskEvent event;
+  private final JavacTypes javacTypes;
   private final Trees trees;
   private final SemanticdbJavacOptions options;
   private final EndPosTable endPosTable;
@@ -42,12 +41,17 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
   private String source;
 
   public SemanticdbVisitor(
-      JavacTask task, GlobalSymbolsCache globals, TaskEvent event, SemanticdbJavacOptions options) {
+      JavacTask task,
+      GlobalSymbolsCache globals,
+      TaskEvent event,
+      SemanticdbJavacOptions options,
+      JavacTypes javacTypes) {
     this.task = task;
     this.globals = globals; // Reused cache between compilation units.
     this.locals = new LocalSymbolsCache(); // Fresh cache per compilation unit.
     this.event = event;
     this.options = options;
+    this.javacTypes = javacTypes;
     this.trees = Trees.instance(task);
     if (event.getCompilationUnit() instanceof JCTree.JCCompilationUnit) {
       this.endPosTable = ((JCTree.JCCompilationUnit) event.getCompilationUnit()).endPositions;
@@ -113,6 +117,7 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
         break;
       case METHOD:
         builder.setKind(Kind.METHOD);
+        builder.addAllOverriddenSymbols(semanticdbOverrides(sym));
         break;
       case CONSTRUCTOR:
         builder.setKind(Kind.CONSTRUCTOR);
@@ -500,6 +505,17 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
     properties |= (sym.flags() & Flags.ABSTRACT) > 0 ? Property.ABSTRACT_VALUE : 0;
     properties |= (sym.flags() & Flags.FINAL) > 0 ? Property.FINAL_VALUE : 0;
     return properties;
+  }
+
+  private List<String> semanticdbOverrides(Symbol sym) {
+    ArrayList<String> overriddenSymbols = new ArrayList<>();
+    Set<Symbol.MethodSymbol> overriddenMethods = javacTypes.getOverriddenMethods(sym);
+
+    for (Symbol.MethodSymbol meth : overriddenMethods) {
+      overriddenSymbols.add(semanticdbSymbol(meth));
+    }
+
+    return overriddenSymbols;
   }
 
   private Semanticdb.Access semanticdbAccess(Symbol sym) {
--- a/tests/unit/src/test/scala/tests/OverridesSuite.scala
+++ b/tests/unit/src/test/scala/tests/OverridesSuite.scala
@@ -1,0 +1,152 @@
+package tests
+
+import scala.meta.Input
+
+import com.sourcegraph.lsif_semanticdb.Symtab
+import munit.FunSuite
+import munit.TestOptions
+
+class OverridesSuite extends FunSuite with TempDirectories {
+
+  val targetroot = new DirectoryFixture()
+
+  override def munitFixtures: Seq[Fixture[_]] =
+    super.munitFixtures ++ List(targetroot)
+
+  def checkOverrides(
+      options: TestOptions,
+      source: String,
+      extractSymbol: String,
+      expectedSymbols: List[String],
+      expectedCount: Int,
+      qualifiedClassName: String = "example.Parent"
+  ): Unit = {
+    test(options) {
+      val compiler = new TestCompiler(targetroot())
+      val relativePath = qualifiedClassName.replace('.', '/') + ".java"
+      val input = Input.VirtualFile(relativePath, source)
+      val result = compiler.compileSemanticdb(List(input))
+      val symtab = new Symtab(result.textDocument)
+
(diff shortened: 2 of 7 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -502,6 +507,17 @@ private int semanticdbSymbolInfoProperties(Symbol sym) {
     return properties;
   }
 
+  private List<String> semanticdbOverrides(Symbol sym) {
```
olafurpg, line 510: I think you can do like this
```suggestion
  private List<String> semanticdbOverrides(Symbol sym) {
    return new ArrayList<>(javacTypes.getOverriddenMethods(sym));
```

olafurpg: Maybe even just return the set directly
```java
private Iterable<String> semanticdbOverrides(Symbol sym) {
  return javacTypes.getOverriddenMethods(sym);
}
```

olafurpg (changes requested):
The fix looks good. Can we add unit tests in `TargetedSuite` that validate that the `overrides` are defined?

It would be good to check the following cases

- override other method that's defined in the same file 
- override other method that's defined outside the file, for example by implementing `AutoCloseable`
- override interface method
- override abstract class method
- override method with type parameters, example
```java
interface Haha<T> {
  void add(T elem);
}
class IntHaha implements Haha<Integer> {
  void add(Integer elem) {}
}
```

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -0,0 +1,157 @@
...
+      qualifiedClassName: String = "example.Parent"
+  ): Unit = {
+    test(options) {
+      val compiler = new TestCompiler(targetroot())
+      val relativePath = qualifiedClassName.replace('.', '/') + ".java"
+      val input = Input.VirtualFile(relativePath, source)
+      val result = compiler.compileSemanticdb(List(input))
+      val symtab = new Symtab(result.textDocument)
+
+      val count = symtab.symbols.get(extractSymbol).getOverriddenSymbolsCount
```
olafurpg, line 29: Another way to write these tests is like this
```suggestion

import scala.jdk.CollectionConverters._
def checkOverrides(
      options: TestOptions,
      source: String,
      extractSymbol: String,
      expected: String
      
     // ...
  val obtained = symtab.symbols.get(extractSymbol).getOverriddenSymbolsList.asScala.mkString("\n")
  assertNoDiff(obtained, expected)
}


checkOverrides(
    "same file",
    """package example;
      |
      |class Parent {
      | public void stuff() {}
      |
      | class Child extends Parent {
      |   @Override
      |   public void stuff() {}
      | }
      |}
      |""".stripMargin,
    "example/Parent#Child#stuff()."
    "example/Parent#stuff()."
  )
   
```

This avoids the repetitive `assertEquals(count, 1) ...`  body for every test case. You can also copy-paste the `""".stripMargin` formatted error message to update the expected value.

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -0,0 +1,157 @@
...
+      |""".stripMargin,
+    "example/Parent#toString().",
+    { case (count, symbols) =>
+      assertEquals(count, 1)
+      assertEquals(symbols.apply(0), "java/lang/Object#toString().")
+    }
+  )
+
+  checkOverrides(
+    "external override",
```
olafurpg, line 72:
```suggestion
    "implement interface",
```

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -0,0 +1,157 @@
...
+      | class IntHaha implements Haha<Integer> {
+      |  void add(Integer elem) {}
+      | }
+      |}
+      |""".stripMargin,
+    "example/Parent#IntHaha#add().",
+    {
+      case (count, symbols) =>
+        assertEquals(count, 1)
+        assertEquals(symbols.apply(0), "example/Parent#Haha#add().")
```
olafurpg, line 110: 👨‍🍳 💋

olafurpg: This is surprisingly difficult to implement if you roll your own semantic analysis

olafurpg (changes requested):
The new test cases look great. One comment on how to make them a bit less repetitive, otherwise this is ready to merge soon!

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
Strum355: `getOverriddenMethods` returns a `Set<Symbol.MethodSymbol>`, hence the need to get the semanticdb symbol string for each

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -0,0 +1,152 @@
...
+
+  override def munitFixtures: Seq[Fixture[_]] =
+    super.munitFixtures ++ List(targetroot)
+
+  def checkOverrides(
+      options: TestOptions,
+      source: String,
+      extractSymbol: String,
+      expectedSymbols: List[String],
+      expectedCount: Int,
```
olafurpg, line 21: This is redundant from `expectedSymbols.length`
```suggestion

```

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -0,0 +1,152 @@
...
+      val result = compiler.compileSemanticdb(List(input))
+      val symtab = new Symtab(result.textDocument)
+
+      val count = symtab.symbols.get(extractSymbol).getOverriddenSymbolsCount
+      val syms = symtab
+        .symbols
+        .get(extractSymbol)
+        .getOverriddenSymbolsList
+        .toArray(new Array[String](count))
+      assertEquals(count, expectedCount)
```
olafurpg, line 37: We can also remove the `zipWithIndez.foreach` assertions
```suggestion
      assertEquals(syms, expected)
```

olafurpg (changes requested):

tests/unit/src/test/scala/tests/OverridesSuite.scala
```diff
@@ -17,29 +18,21 @@ class OverridesSuite extends FunSuite with TempDirectories {
       options: TestOptions,
       source: String,
       extractSymbol: String,
-      expectedSymbols: List[String],
-      expectedCount: Int,
-      qualifiedClassName: String = "example.Parent"
+      expectedSymbols: String *,
```
olafurpg, line 21: It's best to keep this as a single string and pass in multiline strings instead.
```suggestion
       expectedSymbols: String,
```
The motivation for this is that test failures include a multiline formatted string that you can copy-paste back into the source code. If we keep the vararg parameter then you need to manually write out the new expected output or manually split up the failure message somehow.

olafurpg (changes requested):

tests/unit/src/test/scala/tests/OverridesSuite.scala
Strum355: The output is identical, and in both situations theres some manual work required to fix the test with the right expected result, so Im not sure I follow on what the advantage is youre describing. I could see it being better from a visual clarity perspective though, so Ill change it anyways

New:
```scala
checkOverrides(
	// ...
	"""example/Parent#toString().
	|java/lang/Object#toStrig().
	|""".stripMargin
)
```
```diff
==> X tests.OverridesSuite.multiple  0.071s munit.ComparisonFailException: /home/noah/Sourcegraph/lsif-java/tests/unit/src/test/scala/tests/OverridesSuite.scala:34
33:        .getOverriddenSymbolsList.stream.collect(Collectors.joining("\n"))
34:      assertNoDiff(expectedSymbols, syms)
35:    }
diff assertion failed
=> Obtained
    """|example/Parent#toString().
       |java/lang/Object#toStrig().
       |""".stripMargin
=> Diff (- obtained, + expected)
 example/Parent#toString().
-java/lang/Object#toStrig().
+java/lang/Object#toString().
    at munit.FunSuite.assertNoDiff(FunSuite.scala:11)
    at tests.OverridesSuite.$anonfun$checkOverrides$1(OverridesSuite.scala:34)
```

Current:
```scala
checkOverrides(
	// ...
	"example/Parent#toString().",
	"java/lang/Object#toStrig().",
)
```
```diff
==> X tests.OverridesSuite.multiple  0.053s munit.ComparisonFailException: /home/noah/Sourcegraph/lsif-java/tests/unit/src/test/scala/tests/OverridesSuite.scala:35
34:        .getOverriddenSymbolsList.stream.collect(Collectors.joining("\n"))
35:      assertNoDiff(expectedSyms, syms)
36:    }
diff assertion failed
=> Obtained
    """|example/Parent#toString().
       |java/lang/Object#toStrig().
       |""".stripMargin
=> Diff (- obtained, + expected)
 example/Parent#toString().
-java/lang/Object#toStrig().
+java/lang/Object#toString().
    at munit.FunSuite.assertNoDiff(FunSuite.scala:11)
    at tests.OverridesSuite.$anonfun$checkOverrides$1(OverridesSuite.scala:35)
```

olafurpg: The output from a failing test remains identical but the call-site can be updated with a single copy-paste if you use the multiline string syntax. If you use varargs syntax at call-site then you have to post-process the error message.

olafurpg (approved):

tests/unit/src/test/scala/tests/OverridesSuite.scala
Strum355: Not sure Im seeing that in this example :thinking: If multiple lines failed I still need to add `|`/remove `+` or `-`, unless Im missing something here :sweat_smile: 
Either way, code's been updated :slightly_smiling_face:

olafurpg: You can copy-paste the "Obtained" section above the "Diff" section

Strum355: It all makes sense now, I had passed the params the wrong way around to `assertNoDiff` :facepalm:

olafurpg: Gharg! I didn't check. MUnit actually uses the opposite order for actual/expected compared to JUnit (some people are upset about it). The reason the obtained value comes first is because it's prettier to have the multiline string appear as the last argument.

PR #173, opened 2021-04-20
Add support for the Gradle Playframework plugin.

### LINKED ISSUE #170: Index command does nothing in Gradle project with Playframework plugin enabled
```
git clone https://github.com/olafurpg/gradle-example
cd gradle-example
git checkout play-plugin
cs launch lsif-java:0.4.0 --contrib -- index
....
```
The command compiles the project but no LSIF index is generated at the end of it.

olafurpg:
Previously, the `index` command didn't work with Gradle projects that
use the Playframework plugin. The codebase would compile, but no LSIF
index would be created. Now, everything should work as expected.

The problem was that the Playframework plugin uses the Scala plugin to
compile auto-generated template and routes files. The Scala plugin runs
Zinc (Scala incremental compiler) in a daemon process behind the scenes
so it ignores the `javac` fork settings that `lsif-java index` adds to
the build.

The fix is to enable the SemanticDB Java agent on the Zinc daemon
process and to add one more injection point to ensure that the
SemanticDB compiler plugin is always on the classpath for all projects.

Fixes #170

### DIFF
--- a/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/BuildTool.scala
+++ b/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/BuildTool.scala
@@ -16,6 +16,7 @@ abstract class BuildTool(val name: String, index: IndexCommand) {
 
   def indexJdk(): Boolean = true
 
+  final def sourceroot: Path = index.workingDirectory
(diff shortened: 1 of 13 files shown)


Strum355 (approved):
Nice work! :tada:

PR #176, opened 2021-04-21
Remove `abstract`/`static` from interface signature and pretty-print `default` modifier

Strum355:
Previously, top-level interface decls included the `abstract` modifier, while nested interfaces included `abstract static` modifiers. This PR removes those from the pretty-printed signature, while also introducing the `default` modifier to the pretty-printing.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -530,8 +530,12 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
     int properties = 0;
     properties |= sym.isEnum() ? Property.ENUM_VALUE : 0;
     properties |= sym.isStatic() ? Property.STATIC_VALUE : 0;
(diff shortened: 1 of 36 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -530,8 +530,13 @@ private int semanticdbSymbolInfoProperties(Symbol sym) {
     int properties = 0;
     properties |= sym.isEnum() ? Property.ENUM_VALUE : 0;
     properties |= sym.isStatic() ? Property.STATIC_VALUE : 0;
-    properties |= (sym.flags() & Flags.ABSTRACT) > 0 ? Property.ABSTRACT_VALUE : 0;
+    // for default interface methods, Flags.ABSTRACT is also set...
+    properties |=
+        ((sym.flags() & Flags.ABSTRACT) > 0) && ((sym.flags() & Flags.DEFAULT) == 0)
```
olafurpg, line 535: Nit: this condition would read more cleanly on a separate statement
```java
boolean isAbstract = ...
properties |= isAbstract ? Property.ABSTRACT_VALUE : 0;
```

olafurpg (approved):
This is a really nice improvement. LGTM 👍

PR #179, opened 2021-04-22
replace emitting MarkedString[] with MarkupContent

### LINKED ISSUE #162: Emit `MarkupContent` instead of `MarkedString[]` for hover messages
Currently, lsif-java uses `MarkedString[]` to emit hover messages. This format is considered deprecated and we should use `MarkupContent` instead. The benefit of `MarkupContent` is that we should be able to get correct syntax highlighting in the hover messages. Currently, lsif-java hover messages are rendered as monospace with no syntax highlighting.

Strum355:
closes #162 

Required some work in SnapshotLsifCommand.scala due to the new format. Also fixed a load of lint errors that Intellij kept throwing at me :upside_down_face:


lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -174,14 +176,20 @@ object SnapshotLsifCommand {
...
     val hovers: Map[String, String] = monikers.map { case (sym, obj) =>
-      val hoverMessage =
-        for {
-          resultSet <- monikerInverse.get(obj.getId).toList
-          hoverEdge <- hoverEdges.get(resultSet).toList
-          hoverResult <- hoverVertexes.get(hoverEdge).toList
-          contents <- hoverResult.getContentsList.asScala
-        } yield contents.getValue.trim
-      sym -> hoverMessage.mkString("\n\n")
+      val hover =
```
olafurpg, line 179: Nit: for comprehensions become hard to read when you add a complicated chain of operations on top of it. I think this diff would read cleaner by keeping the original for comprehension and adding a separate statement instead

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -174,14 +176,20 @@ object SnapshotLsifCommand {
...
+        (
+          for {
+            resultSet <- monikerInverse.get(obj.getId).toList
+            hoverEdge <- hoverEdges.get(resultSet).toList
+            hoverResult <- hoverVertexes.get(hoverEdge).toList
+            contents <- hoverResult.getContents.getValue.trim
```
olafurpg, line 185: Here you're iterating through every character of the string only to `mkString` it later. You can yield the value directly instead
```suggestion
            } yield hoverResult.getContents.getValue.trim
```

olafurpg: What I would do instead is to replace the enclosing `).mkString(....)` logic with more stuff inside the for comprehension
```scala
        for {
          resultSet <- monikerInverse.get(obj.getId).toList
          hoverEdge <- hoverEdges.get(resultSet).toList
          hoverResult <- hoverVertexes.get(hoverEdge).toList
          contents = hoverResult.getContents.getValue.trim
          codeFence <- contents
            .linesIterator
            .dropWhile(_ != "```java")
            .drop(1)
            .takeWhile(_ != "```")
        } yield codeFence
```

This logic won't crash on an empty string and it only returns an actual result if there is a code fence.

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -209,56 +217,54 @@ object SnapshotLsifCommand {
...
         inputs.getOrElseUpdate(uri, Input.path(Paths.get(URI.create(uri))))
       }
       def renderPos(pos: LsifPosition): String = {
-        s"${pos.getLine()}:${pos.getCharacter()}"
+        s"${pos.getLine}:${pos.getCharacter}"
       }
       val addedVertexes = mutable.Set.empty[String]
       def render(node: LsifObject): String = {
-        (node.getType(), node.getLabel()) match {
+        (node.getType, node.getLabel) match {
```
olafurpg, line 234: You can disable the IntelliJ warning to remove the `()`. I think it's annoying that IntelliJ enables that warning by default. Maybe I should just change Metals to no insert the parentheses by default when calling Java methods. Anyways, no need to change this PR, I'm OK with removing the parentheses

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
```diff
@@ -127,21 +124,27 @@ private Integer processDocument(
...
         if (symbolInformation.hasSignature()) {
-          markedStrings.add(
-              new MarkedString(
-                  doc.semanticdb.getLanguage(),
-                  new SignatureFormatter(symbolInformation, symtab).formatSymbol()));
+          String b =
+              "```"
+                  + doc.semanticdb.getLanguage().toString().toLowerCase(Locale.ROOT)
+                  + '\n'
+                  + new SignatureFormatter(symbolInformation, symtab).formatSymbol()
```
olafurpg, line 139: What do yo uthink about rendering the markup in `SignatureFormatter`? This is a hot path and we should try to to allocate as few intermediary strings as possible. It's best to use only one stringbuilder

olafurpg: Ideally, we want to generate the markup content directly from a single string builder.

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifWriter.java
```diff
@@ -92,21 +92,16 @@ public int emitDefinitionResult(int resultSet) {
...
-                                        .setLanguage(
-                                            ms.language.toString().toLowerCase(Locale.ROOT))
-                                        .setValue(ms.value)
-                                        .build())
-                            .collect(Collectors.toList()))));
+                    .setContents(
+                        Content.newBuilder()
+                            .setKind(markupContents.kind.toString().toLowerCase())
+                            .setValue(markupContents.value)
+                            .build())));
```
olafurpg, line 104:
```suggestion
                            )));
```

olafurpg (changes requested):
Mostly minor comments, otherwise looks good.

lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
Strum355: Genuinely dont know what all this scala stuff does well enough to be able to change it confidently :sweat_smile:

olafurpg: Fair enough. I'm sorry about that, it takes a while to get used to them but once you get the hang of it you can write super concise code and it's difficult to go back. Don't hesitate to ping me directly if you want to update a for comprehension and we can do a walkthrough of how they work.

olafurpg: For comprehensions are awesome because you can write like 5 nested for loops (or nested map/flatMap/filter) as a single flat block.

lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifSemanticdb.java
Strum355: Id be interested in profiling for sure, Im not confident we'll see a notably improvement, given that calling toString on a StringBuilder causes a copy _anyways_

olafurpg: You're probably right. Ignore my comment. I personally try to stick with one stringbuilder per emitted string in hot paths, but I've never validated the performance implications (short-lived allocations on the JVM are surprisingly cheap)

olafurpg (approved):
I took the liberty to push a commit updating the for-comphrension. LGTM 👍

PR #211, opened 2021-05-05
Fix bug when indexing JDK 11 sources

olafurpg:
Previously, PackageHub was unable to auto-index the JDK 11 sources
because it was missing appropriate `--module` flags that are not needed
for JDK 8.


olafurpg:
Superseded by #212

PR #222, opened 2021-05-18
Declare local symbols as "export" instead "import" monikers

olafurpg:
Previously, `localN` symbols were declared as imported monikers. I don't think this change affects anything, but I noticed  this bug while investigating the issue to JDK navigation because I saw lots of `import` monikers in the JDK dump.lsif, when I expected only to see `export` monikers


Strum355:
Should we not emit monikers for `localN` symbols at all instead?

olafurpg:
We currently use local monikers for snapshot testing purposes. Maybe we should have a flag to decide whether to emit local monikers, but this PR is a small enough improvement for now to merit merging

Issue #219, opened 2021-05-14
Allow for `targetroot` to be based off the `javac` directory output


wiwa:
For Pants use it would be helpful to be able to output to a `META-INF` under whatever is passed to `zinc`/`javac`'s output dir (`-d` option).
 
 I played around with a solution in this commit: https://github.com/wiwa/lsif-java/commit/64f7195af92eadf84eaebd1b3f188eb52443efbf 

Not sure if this is how one is supposed to get `javac` options though...

olafurpg:
Thank you for reporting! I am OK with merging that diff so that the targetroot can be automatically configured from the javac output directory. I think it would good to require the user to explicitly enable this behavior with something like `-targeroot:javac-classes-directory`. The targetroot wasn't auto-configurable because I didn't figure out how to extract the default compile directory from the compiler APIs 😅

olafurpg:
The fix has been released as part of https://github.com/sourcegraph/lsif-java/releases/tag/v0.5.3

PR #224, opened 2021-05-18
Allow javac plugin to infer targetroot from javac class output directory

wiwa:
Previously, users were required to specify an absolute path for the `-targetroot` of the `javac` SemanticDB plugin. However, we would like build tools like Pants to be able to output SemanticDB in the same subdirectory as class files since this is how the current Scala SemanticDB plugin behaves (see issue #219)

This change introduces a special argument, `-targetroot:javac-classes-directory`, which will tell the plugin to infer and use the `javac` class output directory, i.e. `javac`'s `-d` argument.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,8 +64,9 @@ commands +=
 
 commands +=
   Command.command("checkAll") { s =>
-    "scalafmtCheckAll" :: "scalafmtSbtCheck" :: "scalafixAll --check" ::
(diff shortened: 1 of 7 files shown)


olafurpg (approved):
Thank you for this contribution! LGTM 👍 I took the liberty to push a commit to add tests so that we ensure this feature doesn't regress in the future :)

PR #225, opened 2021-05-19
Disable auto-indexing on packagehub

olafurpg:
Previously, packagehub automatically scheduled LSIF indexing for repos.
This commit disables that feature so that you need to manually send a
`curl -X POST` requst to trigger LSIF indexing of repos.


Strum355 (approved):
Do we still get indexing of dependencies with this change? 🙂

olafurpg:
You can manually trigger indexing with a `curl -X POST` request, which I what I have been doing. In the long-term pkghub should become obsolete as gitserver supports cloning maven repos and the lsif backend auto-schedules LSIF indexing of `packageInformation` vertexes

PR #227, opened 2021-05-19
Improve handling of moniker import/exports and the JDK

olafurpg:
See individual commits for details. Related issue https://github.com/sourcegraph/sourcegraph/issues/21058

### DIFF
--- a/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/BuildTool.scala
+++ b/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/BuildTool.scala
@@ -13,6 +13,7 @@ abstract class BuildTool(val name: String, index: IndexCommand) {
   protected def defaultTargetroot: Path
 
   def isHidden: Boolean = false
+  def buildKind: String = ""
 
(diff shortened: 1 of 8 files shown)


Strum355:
Just so I understand correctly, will this mean that definitions of jdk classes in non jdk repos will have an import moniker attached to them instead of an export moniker? How does that behaviour work in sourcegraph? Im assuming go-to-def on the "shadowed" definition will bring you to the jdk definition instead, which would seem reasonable, and go-to-def on references to the shadowed symbol will also bring you to the jdk definition instead? The only way to find the shadowed definitions then are via find references. Does this sound right?

olafurpg:
I haven't validated the answers below but will answer with my intuition about the expected behavior.

> Im assuming go-to-def on the "shadowed" definition will bring you to the jdk definition instead, which would seem reasonable, and go-to-def on references to the shadowed symbol will also bring you to the jdk definition instead?

Correct, that's my intuition as well.

> The only way to find the shadowed definitions then are via find references. Does this sound right?

Correct, the shadow definition will only appear as a reference.

Strum355 (approved):

PR #228, opened 2021-05-21
Use "local" instead of "import"  for local symbol monikers.

olafurpg:
Previously, lsif-java used "kind: import" for local symbols. The
Sourcegraph backend stores all import/export monikers in the postgres
database. This change make the kind "local" so that they don't get
stored in the Postgres backed, reducing disk usage.


lsif-java/src/main/scala/com/sourcegraph/lsif_java/commands/SnapshotLsifCommand.scala
```diff
@@ -61,7 +61,7 @@ case class SnapshotLsifCommand(
     } {
       val docPath = AbsolutePath
         .of(Paths.get(doc.getUri), sourceroot)
-        .toRealPath()
+        .normalize()
```
olafurpg, line 64: This change was necessary to get the snapshots testing passing on macos. We don't run the CI on macos, but might want to reconsider that if we hit on more linux/macos issues like this.

tests/snapshots/src/main/generated/index-semanticdb/locals
```diff
@@ -17,41 +17,41 @@ public class Example {
...
+      │ │    │definitionResult(46)│ │  │                      │
+      │ │    ╰──────────┬─────────╯ │  │                      │
+      │ │               │   ╭───────┼──╯                      │
+      │ │               │   │       ╰──────────────────╮      │
+      ╰─┼───────────╮   │   │                          │      │
+        │           │   │   │   ╭──────────────────────┼──────╯
+        │           │   │   │   │                      │
+        v           v   v   │   v                      v
+ ╭────────────╮ ╭───────────┴──────╮ ╭───────────────────────────────────╮
+ │local local0│ │range(50) 3:34 'n'│ │hoverResult(54) ```java\nint n\n```│
```
olafurpg, line 56: You can ignore the rest of the diff in this file, the most interesting bit is `local local0` to indicate we now use the `local` kind for local symbols

efritz (approved):

PR #230, opened 2021-05-24
Tweak ANSI color output 

olafurpg:
Previously, lsif-java used default color output settings when shelling out to Maven and Gradle. Now, lsif-java uses --batch-mode for Maven and `TERM=dumb` for Gradle to reduce ANSI color output (helpful for CI logs). Additionally, we upgrade to the latest moped release so that setting `NO_COLOR=true` disables all ANSI colors (including output from lsif-java).

The motivation for this PR was to reduce noise in the Sourcegraph auto-indexing CI logs, which currently render ANSI escape codes.


Strum355 (approved):
Curious if you wanna disable it only when we set NO_COLOR, so we can still get colour when running in an interactive term 🙂 but thats up to you, just a thought

olafurpg:
@Strum355 I thought about it but it didn't feel worth the development effort. Most users run lsif-java in CI and I think it makes sense to enable `--batch-mode` and `TERM=dumb` by default out of the box

PR #233, opened 2021-05-31
Upgrade to the latest sbt-sourcegraph

olafurpg:
This release uses lsif-java to convert SemanticDB instead of the
outdated lsif-semanticdb implementation.


olafurpg:
I merged this PR right away since the best way to test it is to run the upload step on the main branch. Looks like everything is working correctly now!  Sourcegraph is using hover messages from lsif-java when browsing the sourcegraph/lsif-java repo :D 

image (4)

PR #268, opened 2021-07-22
emit LSIF dump with rw-r--r-- permissions

Strum355:
In development, when docker runs as root (as is default), auto-indexing is not successful as the devenv is running as non-root. This causes LSIF dumps from lsif-java to fail to upload with `error: error: open dump.lsif: permission denied`. In-line with other indexers, this PR makes lsif-java emit the LSIF dump with world-read permissions


olafurpg (approved):
LGTM 👍

olafurpg:
> [warn] lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifWriter.java isn't formatted properly!

Missing `sbt compile` to fix that

PR #308, opened 2021-08-13
change manual gradle configuration to use compileOnly instead of implementation

Strum355:
Due to how Gradle classpath configurations are laid out, we should recommend semanticdb-javac dependency be in the `compileOnly` configuration, as building fatjars by default takes from the `runtimeClasspath` classpath, as well as other runtime implications.

image


olafurpg (approved):
Great catch!

PR #309, opened 2021-08-14
fix range emitted for member references to only include the reference

### LINKED ISSUE #307: Method reference is not broken into sections
Using version: 0.5.7

In the below code in VSCode, clicking anywhere on `this.getThis()::supplyFoo` results in the whole section being highlighted and using `goto` results in navigating to the `supplyFoo` method only.

I'd expect to be able to highlight the individual sections (`this`, `getThis()` and `supplyFoo`) and navigate to the `getThis()` method when I click on that section.

This works fine in method `createFoo` but not in `createFooUsingMethodReference`

I've just assumed this is down to lsif-java, not VSCode/Metals

```java
package Bar;

import java.util.function.Supplier;

public class Foo {

    public Foo supplyFoo() {
        return new Foo();
    }

    public Foo getThis() {
        return this;
    }

    public Foo createFoo() {
        return this.getThis().supplyFoo();  // <- clicking on `supplyFoo` highlights `supplyFoo`
    }

    public Supplier<Foo> createFooUsingMethodReference() {
        return this.getThis()::supplyFoo;   // <- clicking on `supplyFoo` highlights `this.getThis()::supplyFoo`
    }
}
```

Strum355:
Fixes #307 

Unfortunately, the referenced method in a member reference tree node doesn't have a child node. In the example of `this.getThis().getThis().getThis()::supplyFoo`, the only descendant nodes are `this` and each `getThis`. For that reason, this approach uses a basic text search starting from the _end_ position of the member reference tree node.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/CompilerRange.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/CompilerRange.java
@@ -32,15 +32,21 @@ public enum CompilerRange {
 
   /**
    * Use text search to find the start of the symbol name and use (found start + symbol name length)
-   * for the SemanticDB end position;
+   * for the SemanticDB end position.
    */
   FROM_TEXT_SEARCH,
 
   /**
    * Use text search to find the start of the symbol name, using the point position as the starting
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
LGTM 👍 The fix looks good OK given the constraints.

PR #332, opened 2021-09-06
Return non-zero exit code when missing SemanticDB files

olafurpg:
Previously, `lsif-java index` would occasionally return a successful
exit code even if it didn't produce a `dump.lsif` file. Now, it should
return the exit code 1 instead in these scenarios.

### DIFF
--- a/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/LsifBuildTool.scala
+++ b/lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/LsifBuildTool.scala
@@ -160,18 +160,20 @@ class LsifBuildTool(index: IndexCommand) extends BuildTool("LSIF", index) {
       .isDirectory(targetroot.resolve("META-INF"))
     if (errors.nonEmpty && !isSemanticdbGenerated) {
       CommandResult(1, Nil)
-    } else if (errors.nonEmpty && isSemanticdbGenerated) {
-      index
-        .app
-        .reporter
-        .info(
-          "Some SemanticDB files got generated even if there were compile errors. " +
-            "In most cases, this means that lsif-java managed to index everything " +
-            "except the locations that had compile errors and you can ignore the compile errors." +
-            errors.mkString("\n")
-        )
+    } else {
+      if (errors.nonEmpty && isSemanticdbGenerated) {
+        index
+          .app
+          .reporter
+          .info(
+            "Some SemanticDB files got generated even if there were compile errors. " +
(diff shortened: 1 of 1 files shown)


lsif-java/src/main/scala/com/sourcegraph/lsif_java/buildtools/LsifBuildTool.scala
```diff
@@ -160,18 +160,20 @@ class LsifBuildTool(index: IndexCommand) extends BuildTool("LSIF", index) {
...
+            "Some SemanticDB files got generated even if there were compile errors. " +
+              "In most cases, this means that lsif-java managed to index everything " +
+              "except the locations that had compile errors and you can ignore the compile errors." +
+              errors.mkString("\n")
+          )
+      }
+      CommandResult(0, Nil)
     }
-    CommandResult(0, Nil)
   }
```
Strum355, line 178: Im confused about this, does not returning an explicit CommandResult result in the exit code being set to non-zero?

olafurpg: The `CommandResult` on line 162 was previously discarded because the if/else wasn't the last statement

Strum355: oh so thats a thing in scala :thinking: TIL

olafurpg: There are compiler flags to turn that into an error but I don't like using it because there are too many false positives (lots of side-effecting methods return non-Unit values)

olafurpg: I probably messed up because I've been writing too much Java/Go lately with early returns. I don't remember being hit by this in the past, it's usually a non-issue in my experience.

Strum355: It mightve been introduced by me [here](https://github.com/olafurpg/lsif-java/commit/928dbdf651808afc91b9d062485582855833a560#diff-98e378ae471029019ae0c8869f19885a0d7e5e297363ee4a22d8ff59a9381663)

olafurpg: Aha, hehe looks like that indeed 😜

Strum355 (approved):

PR #361, opened 2021-10-26
Add support for Protobuf in Gradle projects

olafurpg:
Previously, lsif-java only indexed the generated code from proto files.
Now, we also index the original proto files so that it's possible to
navigate from Java/Scala/Kotlin to Protobuf (and vice-versa).

This new feature only works for Gradle at the moment but the underlying
approach is generic and can be reused for other build tools as well.

Towards https://github.com/sourcegraph/lsif-java/issues/363

TODOs:

- [x] nested types
- [x] `java_package`
- [x] `java_multiple_files`
- [x] `java_outer_classname`
- [x] `map`
- [x] `message`
- [x] `repeated`
- [x] `enum`
- [x] `string`
- [x] `double`
- [x] `float`
- [x] `int32`
- [x] `int64`
- [x] `uint32`
- [x] `uint64`
- [x] `sint32`
- [x] `sint64`
- [x] `fixed32`
- [x] `fixed64`
- [x] `sfixed32`
- [x] `sfixed64`
- [x] `bool`
- [x] `bytes`
- [ ] `oneof`
- [ ] `service`
- [ ] `rpc`
- [ ] `google.protobuf.Any`
- [ ] `optimize_for = SPEED`
- [ ] `optimize_for = CODE_SIZE`
- [ ] `optimize_for = LITE_RUNTIME`
- [ ] Gradle integration test


olafurpg:
We have validated that the overall approach is good but we want to complete the following upstream issue before shipping Protobuf support. Closing this PR until we pick this up again https://github.com/sourcegraph/sourcegraph/issues/27403

PR #371, opened 2021-11-22
Bump to the latest lsif-kotlin

olafurpg:


tests/snapshots/src/main/generated/com/airbnb/epoxy/ActivityRecyclerPool.kt
```diff
@@ -38,44 +38,44 @@ import java.util.ArrayList
...
+//              ^^^^^ definition com/airbnb/epoxy/ActivityRecyclerPool#pools.  private final val pools: java.util.ArrayList<com.airbnb.epoxy.PoolReference>
+//              ^^^^^ definition com/airbnb/epoxy/ActivityRecyclerPool#getPools().  private final val pools: java.util.ArrayList<com.airbnb.epoxy.PoolReference>
 //                      ^^^^^^^^^ reference java/util/ArrayList#`<init>`().
 //                                ^^^^^^^^^^^^^ reference com/airbnb/epoxy/PoolReference#
 
     @JvmOverloads
 //   ^^^^^^^^^^^^ reference kotlin/jvm/JvmOverloads#`<init>`().
     fun getPool(
-//      ^^^^^^^ definition com/airbnb/epoxy/ActivityRecyclerPool#getPool().
+//      ^^^^^^^ definition com/airbnb/epoxy/ActivityRecyclerPool#getPool().  public final fun getPool(context: [ERROR : Context], poolFactory: () -> [ERROR : RecyclerView.RecycledViewPool]): com.airbnb.epoxy.PoolReference
```
olafurpg, line 57: FYI @Strum355 the Kotlin signature renderer uses "ERROR" when the code has compile errors. I'm not sure what the best behavior is, I personally prefer it like this over returning an empty string. It's easier to report a bug if the signature has "ERROR". This problem happens because the Android SDK is not on the classpath.

PR #392, opened 2022-01-11
fix MD5 checksum computation

### LINKED ISSUE #391: Wrong md5 sum returned
When semanticdb is run on a file: 

```
package a;
public class Main {
  abstract class A {
      abstract void hello();
  }
  abstract class B extends A{
      @Override void hello(){
          System.out.println("Hello!");
      }
  }
  class C extends B{
      @Override void hello(){
          System.out.println("Bye!");
      }
  }
}
```

Java semanticdb returns:

 `D41D8CD98F00B204E9800998ECF8427E` 

while running `md5sum Main.java` returns:

`9F1F9093CB007C3D3A57BA6D7C871858`

This is also the same as returned by https://github.com/scalameta/metals/blob/main/mtags/src/main/scala/scala/meta/internal/mtags/MD5.scala

Strum355:
The CharBuffer dance added about 30 null bytes to the then hashed file data, causing an incorrect MD5 checksum. This PR avoids this erroneous codepath to give us the correct checksum. Closes #391 

image

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/MD5.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/MD5.java
@@ -26,10 +26,7 @@ public final class MD5 {
   }
 
   public static String digest(CharSequence chars) throws NoSuchAlgorithmException {
-    CharBuffer buf = CharBuffer.wrap(chars);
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
Thank you so much for fixing this 🙏

PR #400, opened 2022-02-24
Fix code navigation for synthetic Scala symbols (including case classes)

### LINKED ISSUE #399: Case Class "Find References" Finds All Case Class Definitions
`lsif-java` version: `0.7.5`

Type of code: case classes in Scala

Expected behavior: Case class "Find Reference" should only include usage of the case class in precise code intelligence.

Actual behavior: Case class "Find References" includes ALL case class definitions in precise code intelligence. This could have been due to the fix in https://github.com/sourcegraph/lsif-java/issues/396 introducing this new bug.

```
case class TestA(a: String) {}

case class TestB(a: String) {}
```

olafurpg:
Previously, doing "find references" on case classes showed usages of all
case classes because lsif-java emitted `next` edges for synthetic
symbols. Now, synthetic symbols have asymmetric navigation, goto
definition works from call-site but "find references" does not work at
the definition site.

Fixes #399

### DIFF
--- a/lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
+++ b/lsif-java/src/main/scala/com/sourcegraph/lsif_java/SemanticdbPrinters.scala
@@ -4,11 +4,11 @@ import scala.jdk.CollectionConverters._
 
 import com.sourcegraph.lsif_java.commands.CommentSyntax
 import com.sourcegraph.lsif_java.commands.SnapshotLsifCommand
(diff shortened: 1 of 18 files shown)


olafurpg:
Live example demonstrating the bugfix https://sourcegraph.com/github.com/sourcegraph/sbt-sourcegraph@df4c9f22919bf9843c12afbd9a65f74a3e4b1dfd/-/blob/example/foo.scala?L2:12

PR #404, opened 2022-02-25
Promote synthetic definitions into normal definitions if they share names

### LINKED ISSUE #403: Case Class References Missing Instantiations
`lsif-java` version: `0.7.6`

Type of code: case classes in Scala

Expected behavior: Case class "Find Reference" should contain instantiations.

Actual behavior: Case class "Find References" does not contain instantiations. You can find other instantiations if you "Find Reference" any one instantiation.

If the companion object is explicitly defined, you can find instantiations there by "Find Reference".

Note that "Go to Definition" from instantiations does go to the case class if the companion object does not exist (otherwise, it would go to the companion object), but the hover references the companion object(???). Ideally, all the synthetic functions should reference the case class (and vice versa) similar to IntelliJ's code reference.

```
case class TestA(a: String) {}  // "Find References" missing the below instantation

...

TestA("blah")  // Hover references object TestA but "Go to Definition" jumps to case class definition
```

olafurpg:
Fixes #403 

Previously, we generated synthetic definitions for all synthetic
definitions. However, this meant that "find references" never showed
results for synthetic symbols share the same name as the non-synthetic
definition. This commit changes that so synthetic definitions that share
the same name as the non-synthetic symbol are included with "find
references" results.

While testing this change I noticed that we don't handle synthetic
constructor parameters so I added support for those along the way.

### Test plan

Snapshot tests included in the diff. I also tested the change manually in this URL here https://sourcegraph.com/github.com/sourcegraph/sbt-sourcegraph@e2f71270ea518766c65097b1f17e97b55059d792/-/blob/example/foo.scala?L3:13#tab=references

CleanShot 2022-02-25 at 15 20 44@2x
CleanShot 2022-02-25 at 15 20 47@2x

CleanShot 2022-02-25 at 15 20 51@2x

### DIFF
--- a/lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifTextDocument.java
+++ b/lsif-semanticdb/src/main/java/com/sourcegraph/lsif_semanticdb/LsifTextDocument.java
@@ -98,13 +98,14 @@ public class LsifTextDocument {
       if (definition != null) {
         continue;
       }
-      for (String alternativeSymbol : alternativeSymbols(info)) {
+      for (Semanticdb.SymbolOccurrence alternativeSymbol : alternativeSymbols(info)) {
         Semanticdb.SymbolOccurrence alternativeDefinition =
-            definitionOccurrences.get(alternativeSymbol);
+            definitionOccurrences.get(alternativeSymbol.getSymbol());
         if (alternativeDefinition != null) {
           builder.addOccurrences(
-              Semanticdb.SymbolOccurrence.newBuilder(alternativeDefinition)
-                  .setRole(Semanticdb.SymbolOccurrence.Role.SYNTHETIC_DEFINITION)
+              Semanticdb.SymbolOccurrence.newBuilder()
+                  .setRange(alternativeDefinition.getRange())
+                  .setRole(alternativeSymbol.getRole())
                   .setSymbol(info.getSymbol()));
           break;
         }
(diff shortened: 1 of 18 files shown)


olafurpg:
@bobheadxi The pr-auditor check failed with a cryptic error message https://github.com/sourcegraph/lsif-java/runs/5334788397?check_suite_focus=true#step:4:34
```
![CleanShot 2022-02-25 at 15 20 51@2x](https://user-images.githubusercontent.com/1408093/155731053-74904d20-db13-4e49-b25a-2ebe55f1f41f.png) Error:<nil>}
[35](https://github.com/sourcegraph/lsif-java/runs/5334788397?check_suite_focus=true#step:4:35)
2022/02/25 14:23:17 preMergeAudit: CreateStatus: POST https://api.github.com/repos/sourcegraph/lsif-java/statuses/0c51ae30ffd4e50b1f5fd0a2dbf47e039a3c6b09: 401 Bad credentials []
```

Any advice on how to troubleshoot that?

bobheadxi:
@olafurpg we use `sourcegraph-bot` to make API requests, does the bot have access to this repo?

bobheadxi:
Ah I see - @olafurpg this PR seems to be from a fork, I'm not sure how API access to set status updates on a fork's commits works

wchau:
When I hover

`B`

in

`case class B`

It sometimes gives "case class B..." and sometimes gives "object B".

olafurpg:
@wchau I'm able to reproduce! That's an issue outside of lsif-java that needs to be fixed in the Sourcegraph backend. Both the case class and the object are defined at the same location and there appears to b non-determinism related to which symbol docstring gets displayed in the hover. I've added a reminder to report an upstream issue in the sourcegraph/sourcegraph erpo

olafurpg:
> Ah I see - @olafurpg this PR seems to be from a fork, I'm not sure how API access to set status updates on a fork's commits works

I can open PRs from the main repo but what should the pr-auditor do for PRs from external contributors? I have a preference for opening PRs from my fork instead of the main repo because I like push a lot of WIP branches while keeping the main repo clean. I don't like cloning a repo with lots of WIP branches from the maintainers.

varungandhi-src (approved):

olafurpg:
@wchau I opened a followup issue to address the non-deterministic hovers https://github.com/sourcegraph/sourcegraph/issues/32044

Issue #422, opened 2022-04-06
LSIF fat jar index fails on windows


shemuwel:
Hello.

I've been following the guide presented here: https://sourcegraph.github.io/lsif-java/docs/getting-started.html (Windows environment). However when running `lsif-java index` on my gradle project (e.g. https://github.com/Itiviti/gradle-msbuild-plugin) the path constructed in init-script.gradle & java-toolchain.gradle doesn't seem valid:
```
$ lsif-java index
$ D:\gradle-msbuild-plugin\gradlew.bat --init-script C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\java-toolchains.gradle lsifDetectJavaToolchains

FAILURE: Build failed with an exception.

* Where:
Initialization script 'C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\java-toolchains.gradle' line: 4

* What went wrong:
Could not compile initialization script 'C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\java-toolchains.gradle'.
> startup failed:
  initialization script 'C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\java-toolchains.gradle': 4: unexpected char: '\' @ line 4, column 32.
         java.nio.file.Paths.get('C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\gradle-version.txt'),
                                    ^

  1 error


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 811ms
$ D:\Work\2_GITHUB\gradle-msbuild-plugin\gradlew.bat --no-daemon --init-script C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\init-script.gradle clean lsifJavaDependencies
To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/6.9/userguide/gradle_daemon.html#sec:disabling_the_daemon.
Daemon will be stopped at the end of the build

FAILURE: Build failed with an exception.

* Where:
Initialization script 'C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\init-script.gradle' line: 16

* What went wrong:
Could not compile initialization script 'C:\Users\usr\AppData\Local\Temp\lsif-java17864616560350844221\init-script.gradle'.
> startup failed:
[...]

PR #435, opened 2022-05-03
Run CI on Windows

olafurpg:
Previously, we didn't test lsif-java on Windows resulting in bugs like
https://github.com/sourcegraph/lsif-java/issues/422. This commit adds a
build matrix to our CI job and fixes path escaping logic so that it
should work correctly on Windows.

### Test plan

See new CI job for Windows.


olafurpg:
Superseded by #436 which is opened from origin instead of my fork.

PR #439, opened 2022-05-05
Add Bazel support, fixes #88

### LINKED ISSUE #88: Improved support for Bazel
Related discussions here https://github.com/sourcegraph/sourcegraph/issues/2982#issuecomment-1006670213

At the moment: Bazel requires manual configuration, whereas Maven (for Java) and Gradle (for Java, Scala and Kotlin) just work out-of-the-box without manual configuration.

As per the linked comment:

> Bazel is unlikely to get the same out-of-the-box support as Gradle and Maven because that would eliminate many of the biggest benefits of Bazel: distributed builds and reliable caching

There are a couple of potential options mentioned in that comment; using a `config_setting` or auto-generating `semanticdb` targets. We should:
- [ ] Validate which approach works (or if both work).
- [ ] Add docs on how to set up and maintain lsif-java indexing with Bazel.
- [ ] Document pros and cons of the approach we recommend.

olafurpg:
This PR adds a new Bazel build to this codebase so that Bazel users can
build the semanticdb-javac compiler plugin from source and use it in
their Bazel build.  This PR does not yet include documentation  on how to use this configuration,
but this PR adds an end-to-end example of how to import the lsif-java rules via `include(...)`

### Test plan

See new "bazel" CI job.


varungandhi-src:
> Manually test the Bazel build. We don't yet compile the Bazel build in CI.

How hard would it be to add this?

WORKSPACE
```diff
@@ -0,0 +1,43 @@
...
+##########
+http_archive(
+    name = "rules_proto",
+    sha256 = "66bfdf8782796239d3875d37e7de19b1d94301e8972b3cbd2446b332429b4df1",
+    strip_prefix = "rules_proto-4.0.0",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/rules_proto/archive/refs/tags/4.0.0.tar.gz",
+        "https://github.com/bazelbuild/rules_proto/archive/refs/tags/4.0.0.tar.gz",
+    ],
+)
```
varungandhi-src, line 14: I was looking at the `rules_proto` repo and this release seems to be from Sep 15, 2021. There have been newer releases since then, such as one on Apr 05, 2022 (however, that has a version number '4.0.0-3.20.0' which I don't know how to interpret). Is there a specific reason for picking the older release?

olafurpg:
I'm working on expanding this PR to include full support. We can keep this PR open until all the pieces are ready.

maven_install.json
```diff
@@ -0,0 +1,33 @@
+{
```
varungandhi-src, line 1: suggestion(non-blocking): Could you add a [linguist-generated entry](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github) for this file?

semanticdb-javac/BUILD
```diff
@@ -0,0 +1,41 @@
...
+config_setting(
+    name = "is_enabled",
+    flag_values = {":enabled": "true"},
+)
+
+string_flag(
+    name = "enabled",
+    values = ["true", "false"],
+    build_setting_default = "false",
+)
```
varungandhi-src, line 17: What's the reason for having both `is_enabled` and `enabled`? 😅

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbJavacOptions.java
```diff
@@ -23,6 +23,7 @@ public class SemanticdbJavacOptions {
   public boolean includeText = false;
   public boolean verboseEnabled = false;
   public final ArrayList<String> errors;
+  public boolean isErrorReported = false;
```
varungandhi-src, line 26:
```suggestion
  public boolean alreadyReportedErrors = false;
```

(Or maybe `finishedReportingErrors`.)

Right now, the wording makes it a little ambiguous without looking at the usage site. When I first read this, I was wondering if it meant "should we report errors at all" or "have errors been reported already."

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbJavacOptions.java
```diff
@@ -73,7 +78,11 @@ public static SemanticdbJavacOptions parse(String[] args, Context ctx) {
     if (result.targetroot == null && !useJavacClassesDir) {
       result.errors.add(missingRequiredDirectoryOption("targetroot"));
     }
-    if (result.sourceroot == null) {
+    if (result.uriScheme != UriScheme.BAZEL && result.sourceroot == null) {
```
varungandhi-src, line 81: Would it make sense to create a small function/static method for this check:

```
result.uriScheme != UriScheme.BAZEL && result.sourceroot == null
```

It looks like the negation of this is (correctly) tested in the `inferBazelSourceroot` method, having a common name would make the connection immediately clear.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -91,13 +112,60 @@ public static Path absolutePathFromUri(SemanticdbJavacOptions options, URI uri)
...
+    // /home/repo/com/example/Hello.java
+    // uriPath is the sandbox/temporary file path, for example
+    // /private/var/tmp/com/example/Hello.java
+    //
+    // We infer sourceroot by iterating the names of both files in reverse order and stop at the
+    // first entry where  the two paths are different. For the  example above, we compare
+    // "Hello.java", then "example",  then "com",  and when we reach  "repo" !+ "tmp" then we guess
+    // that "/home/repo" is the sourceroot. This  logic is brittle  and it would be nice to use more
+    // dedicated APIs, but Bazel actively makes an effort to  sandbox compilation  and hide access
+    // to the original workspace, which is why we resort to  solutions like this.
```
varungandhi-src, line 149: nit: This comment has inconsistent use of spaces (bunch of double spaces)

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -91,13 +112,60 @@ public static Path absolutePathFromUri(SemanticdbJavacOptions options, URI uri)
...
+    int uriPathDepth = uriPath.getNameCount();
+    int absolutePathDepth = absolutePath.getNameCount();
+    while (relativePathDepth < uriPathDepth && relativePathDepth < absolutePathDepth) {
+      String uriName = uriPath.getName(uriPathDepth - relativePathDepth - 1).toString();
+      String pathName = absolutePath.getName(absolutePathDepth - relativePathDepth - 1).toString();
+      if (!uriName.equals(pathName)) {
+        break;
+      }
+      relativePathDepth++;
+    }
```
varungandhi-src, line 160: suggestion(non-blocking): Perhaps add some unit tests for this function or specifically this logic? It seems a bit tricky.

WORKSPACE
olafurpg: No specific reason. We only use basic features so I suspect it doesn't matter what version we use. Either way, I updated to the latest version and added a comment explaining how to keep the version updated.

maven_install.json
olafurpg: Good idea, done.

semanticdb-javac/BUILD
olafurpg: One of them defines the name when specifying the flag from the command-line  `bazel build //... //semanticdb-javac:enabled=true`.

The  `is_enabled` target is used as a filter for a `select(...)` clause.

See official docs https://bazel.build/docs/configurable-attributes

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbJavacOptions.java
olafurpg: Good point, applied suggestion. I default to using `is` as the prefix to boolean fields but in this case I agree that `alreadyReportedErrors` is clearer.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbJavacOptions.java
olafurpg: Good point. I refactored this condition into a separate method and added comments to clarify the behavior
```java
  private static boolean isSourcerootDefined(SemanticdbJavacOptions options) {
    if (options.uriScheme == UriScheme.BAZEL) {
      return true; // The sourceroot is automatically inferred for Bazel.
    }
    return options.sourceroot != null;
  }
```

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
olafurpg: Good catch. The Java formatter we're using line-wraps comments, and in my attempts to manually line-wrap I added spaces while joining lines.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
olafurpg: Good suggestion. I opened a reminder issue and a `// FIXME https://github.com/sourcegraph/lsif-java/issues/444` comment. Don't want to delay this PR further.

olafurpg (commented):
Thank you @varungandhi-src for the detailed review! I appreciate the thoughtful comments and the resulting code is much cleaner.

maven_install.json
olafurpg: I removed the maven_install.json files since dependency pinning was causing CI failures and I don't think we need to pin these files, the resolution should always be stable anyways.

varungandhi-src (approved):
LGTM, except for a couple of pending minor comments.

PR #458, opened 2022-06-09
Use a better Java package name for generated Protobuf classes

olafurpg:
### Test plan
See the CI go green.


varungandhi-src (approved):

olafurpg (commented):
It may seem problematic that this repo has a modified copy of the original scip.proto but it's not really an issue IMO. We can fix more stuff later but right I just want to stop generating code to the `lib.codeintel.scip` package.

PR #464, opened 2022-06-24
Include missing `--add-exports` flags for Java 9+ Gradle builds

### LINKED ISSUE #463: Automatically add --add-exports arguments for forked Java compilation in Gradle
Following up from https://sourcegraph.slack.com/archives/CHXHX7XAS/p1655993697067219?thread_ts=1655992386.346109&cid=CHXHX7XAS

We should consider using something like the approach listed here https://github.com/tbroyer/gradle-errorprone-plugin/issues/57#issuecomment-803573280 to automatically add missing `--add-exports` flags for Java 17.

olafurpg:
Previously, we didn't include the necessary `--add-exports` flags when
indexing Java 17 builds with Gradle that don't use toolchains. This
commit fixes that bug so we can correctly index repos like
airbytehq/airbyte.

Fixes #463 

### Test plan

Manually tested this by indexing repos locally. We have pretty good infrastructure to write e2e integration tests but this one requires changing the infrastructure because we need to run our test suite itself with Java 17 to reproduce it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         # NOTE(olafurpg) Windows is not enabled because it times out due to reasons I don't understand.
         # os: [windows-latest, ubuntu-latest]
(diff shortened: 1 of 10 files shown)


varungandhi-src:
This PR was mentioned in a comments on this [StackOverflow question](https://stackoverflow.com/questions/72730404/indexing-a-java-17-project-using-sourcegraphs-scip-java).

PR #476, opened 2022-08-02
feat: add support for the Mill build tool

ckipp01:
This PR adds in support for the Mill build tool by utilizing
https://github.com/ckipp01/mill-scip. It's a pretty thin wrapper. You
can find more details about mill-scip in the repo but the general
workflow is:

- It's an external Mill plugin that gathers everything needed to do a
  compile via Mill, but then adds some extra plugins/scalacOptions to
  ensure semanticDB gets produced
- Once produced it uses `scip-java` as a library to actually produce the
  scip file.

The plugin isn't actually a dependency here which ensures we don't have
any cyclical dependency on anything, but rather just utilizes the
`--import` functionality of Mill and external plugins. I'll leave a
couple more comments on various parts of the PR.

### Test plan
There is a test for the Mill integration

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/MillBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/MillBuildTool.scala
@@ -1,0 +1,110 @@
+package com.sourcegraph.scip_java.buildtools
+
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
+
+import scala.jdk.CollectionConverters._
+
+import com.sourcegraph.scip_java.commands.IndexCommand
+import com.sourcegraph.scip_java.BuildInfo
+
+class MillBuildTool(index: IndexCommand) extends BuildTool("mill", index) {
+
+  override def usedInCurrentDirectory(): Boolean =
+    Files.isRegularFile(index.workingDirectory.resolve("build.sc"))
+
+  override def generateScip(): Int =
+    millVersion() match {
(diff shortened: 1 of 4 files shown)


tests/buildTools/src/test/scala/tests/MillBuildToolSuite.scala
```diff
@@ -0,0 +1,32 @@
+package tests
+
+class MillBuildToolSuite extends BaseBuildToolSuite {
```
ckipp01, line 3: So this doesn't pass right now because to my understanding the tests rely on being able to pass in `targetRoot` which `mill-scip` doesn't support. I didn't really see a reason to when creating it, but I guess I could add it if it's a must to be used here. Let me know, and I'll hold off until getting feedback.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/MillBuildTool.scala
```diff
@@ -0,0 +1,110 @@
...
+  private def failFast(message: String): Int = {
+    index.app.error(message)
+    1
+  }
+
+  private def unconditionallyGenerateScip(): Int = {
+    val localMill = Files.isRegularFile(millFile)
+    val command =
+      if (localMill) {
+        "./mill"
```
ckipp01, line 42: This is also one thing I need to test first locally to ensure proc'ing like this with `./mill` actually works. I figured it's a good idea to try and just use the local mill if they have one.

ckipp01:
Sorry for the noise, created this on my main so closing in favor of #477

Issue #306, opened 2021-08-13
Mill support


olafurpg:
Currently, the `lsif-java index` command only works with Gradle, Maven and sbt. It would be nice to support [Mill](https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html) builds as well.

ckipp01:
This should be able to be closed now as of #477

olafurpg:
Great! Thank you @ckipp01 🙏

PR #477, opened 2022-08-02
feat: add support for the Mill build tool

ckipp01:
This PR adds in support for the Mill build tool by utilizing
https://github.com/ckipp01/mill-scip. It's a pretty thin wrapper. You
can find more details about mill-scip in the repo but the general
workflow is:

- It's an external Mill plugin that gathers everything needed to do a
  compile via Mill, but then adds some extra plugins/scalacOptions to
  ensure semanticDB gets produced
- Once produced it uses `scip-java` as a library to actually produce the
  scip file.

The plugin isn't actually a dependency here which ensures we don't have
any cyclical dependency on anything, but rather just utilizes the
`--import` functionality of Mill and external plugins. I'll leave a
couple more comments on various parts of the PR.

_NOTE_ This only currently supports Scala and Scala 3 projects. There
are some issues with Java projects which will need some upstream work.
You can track that in https://github.com/com-lihaoyi/mill/issues/1983.

Refs #306


tests/buildTools/src/test/scala/tests/MillBuildToolSuite.scala
```diff
@@ -0,0 +1,32 @@
+package tests
+
+class MillBuildToolSuite extends BaseBuildToolSuite {
```
ckipp01, line 3: So this doesn't pass right now because to my understanding the tests rely on being able to pass in targetRoot which mill-scip doesn't support. I didn't really see a reason to when creating it, but I guess I could add it if it's a must to be used here. Let me know, and I'll hold off until getting feedback.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/MillBuildTool.scala
```diff
@@ -0,0 +1,110 @@
...
+  private def failFast(message: String): Int = {
+    index.app.error(message)
+    1
+  }
+
+  private def unconditionallyGenerateScip(): Int = {
+    val localMill = Files.isRegularFile(millFile)
+    val command =
+      if (localMill) {
+        "./mill"
```
ckipp01, line 42: This is also one thing I need to test first locally to ensure proc'ing like this with ./mill actually works. I figured it's a good idea to try and just use the local mill if they have one.

ckipp01: Was able to verify this worked:

```
❯ cs launch com.sourcegraph:scip-java_2.13:0.8.2-1-53a602d2-SNAPSHOT -- index
$ ./mill --import ivy:io.chris-kipp::mill-scip::0.2.1 io.kipp.mill.scip.Scip/generate
Compiling /Users/ckipp/Documents/scala-workspace/mill-scip/build.sc
[1/1] io.kipp.mill.scip.Scip.generate
Found 1 total modules to process.
[1/1] io.kipp.mill.scip.Scip.generate > [12/44] plugin.generatedSources.overridden.mill.scalalib.JavaMo
[1/1] io.kipp.mill.scip.Scip.generate > [20/44] plugin.compileIvyDeps.overridden.mill.scalalib.JavaModu
[1/1] io.kipp.mill.scip.Scip.generate > [23/44] plugin.ivyDeps.overridden.mill.scalalib.JavaModule.ivyD
[1/1] io.kipp.mill.scip.Scip.generate > [25/44] plugin.mandatoryIvyDeps.overridden.mill.scalalib.JavaMo
[1/1] io.kipp.mill.scip.Scip.generate > [33/44] plugin.scalacPluginIvyDeps.overridden.mill.scalalib.Sca
[1/1] io.kipp.mill.scip.Scip.generate > [36/44] plugin.scalacOptions.overridden.mill.scalalib.ScalaModu
[1/1] io.kipp.mill.scip.Scip.generate > [37/44] plugin.scalacOptions.overridden.com.goyeau.mill.scalafi
[1/1] io.kipp.mill.scip.Scip.generate > [43/44] mill.scalalib.ZincWorkerModule.worker
Ensuring everything needed for semanticdb version [4.5.11] is available
Generating semanticdb for [mill-scip_mill0.10_2.13]
[info] compiling 5 Scala sources to /Users/ckipp/Documents/scala-workspace/mill-scip/out/io/kipp/mill/scip/Scip/generate.dest/classes ...
[info] Ignoring duplicate plugin semanticdb (scala.meta.internal.semanticdb.scalac.SemanticdbPlugin)
[info] done compiling
Creating a index.scip in /Users/ckipp/Documents/scala-workspace/mill-scip/out/io/kipp/mill/scip/Scip/generate.dest/index.scip
Including 130 classpath entries
Staring to create a scip index from 5 SemanticDB documents.
[1/1] io.kipp.mill.scip.Scip.generate | [5/5] Finished creating your scip index
info: /Users/ckipp/Documents/scala-workspace/mill-scip/index.scip
```

ckipp01:
Gentle ping on this @olafurpg in case you're not watching the PRs. I'm assuming you're the one that will review this.

olafurpg (approved):
This looks great! Thanks for the ping. I've been on vacation and I didn't see this PR until now 😅

olafurpg:
The CI is failing because `mill` is not installed on the system. Ideally, I'd prefer if the test case included a bootstrapping `mill` script in the root of the generated repo so that users can `sbt test` without having to pre-install `mill`. @ckipp01 could you look into this? I recall that Mill has some bootstrapping script but I'm not familiar with how to install it.

You can BTW run arbitrary code to copy the bootstrap script through the `initCommand =` parameter of the `checkBuild(...)` method. See `GradleBuildToolSuite` for how it's used to create a Gradle wrapper script.

ckipp01:
> The CI is failing because `mill` is not installed on the system. Ideally, I'd prefer if the test case included a bootstrapping `mill` script in the root of the generated repo so that users can `sbt test` without having to pre-install `mill`. @ckipp01 could you look into this? I recall that Mill has some bootstrapping script but I'm not familiar with how to install it.
> 
> You can BTW run arbitrary code to copy the bootstrap script through the `initCommand =` parameter of the `checkBuild(...)` method. See `GradleBuildToolSuite` for how it's used to create a Gradle wrapper script.

Ahh yea of course, I didn't think of that. I went ahead and added the mill script in test resources and will not copy it over. However the tests will still fail. From my comment up above:

> So this doesn't pass right now because to my understanding the tests rely on being able to pass in targetRoot which mill-scip doesn't support. I didn't really see a reason to when creating it, but I guess I could add it if it's a must to be used here. However I notice the sbt test sufffers from the same thing I think. Let me know what you think.

So in the test output you can see it actually finds the SemanticDB and creates the index, but the tests end up looking in the wrong place for it.

ckipp01:
Actually, I was just able to add in a `targetRoot` to `checkBuild`. Now we'll default to what we were doing before unless it's set, and then we'll look there. We sort of cheat this in Mill not because Mill cares that we're passing it it, but instead the test knows where to look for the packages and semanticdb.

olafurpg:
Released scip-java v0.8.4 yesterday and was able to `scip-java index` the coursier/coursier repo thanks to this contribution! Very cool 😄

PR #483, opened 2022-08-16
refactor(build): remove explicit dependency on Bloop

ckipp01:
This is a bit opinionated, so no worries if you'd rather not, but when
working on #477 I had a pretty hard time with Bloop, but when using sbt
things seems to go much smoother. This pr does a couple things:

  - Updates the contributing docs from "don't use <insert-editor>" ->
    Feel free to use Metals with lesser support than IntelliJ will give
    you. It also then recommends using sbt as your build server.
  - Remove the explicit dependency on `sbt-bloop`. This both gets rid of
    some complexity in the build by removing the Bloop-explicit stuff
    which I assume most non-Scala devs aren't familiar with.
  - Removes the `BuildInfo.bloopVersion` which didn't seem to be used
    anywhere.

### Test plan

All tests should remain 🟢 .


olafurpg (approved):
Thank you!

PR #484, opened 2022-08-16
chore(ci): align usage of setup-java

ckipp01:
I saw in one place this was using setup-java but the rest setup-scala.
Another opinionated change, but since setup-java also has the default
sbt launcher and built-in sbt caching I think it's a better choice to
just use that unless it's being included for another reason.

This also makes a small change on the sourcegraph upload to use the npm
installed src instead of curling it down manually.

### Test plan

Everything should remain 🟢 

Supersedes #454


.github/workflows/sourcegraph.yml
```diff
@@ -3,24 +3,25 @@ on:
...
   scip:
     runs-on: ubuntu-latest
     name: "Upload SCIP"
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v13
+      - uses: actions/setup-java@v3
       - uses: actions/setup-go@v2
         with:
           go-version: "1.15.6"
```
ckipp01, line 18: Is go actually needed? I know in my uploads I don't include this and it seems to work fine?

olafurpg: This might be a remnant from the time when we build `src` from source. If CI is green without this step then it's fine to remove it

olafurpg (approved):
This is fantastic! Thank you so much.

.github/workflows/sourcegraph.yml
ckipp01: 👍🏼 I'll send in another pr after this one and double-check. I don't want to edit this one again because I opened it before my first pr was merged, so you'll have to keep clicking run 😆 . If I open another I should be able to see the result and change if necessary.

PR #497, opened 2022-08-25
feat: Mill improvements

ckipp01:
This pr does a few different things to better support Mill. Each commit has a
full description, so it's best to view each of those to get the full idea.

### Test plan

There are new tests added to show that Java module support works, new tests to
show that you can now pass `--output dump.lsif` when using Mill, and the rest of
the tests should stay passing.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/UriScheme.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/UriScheme.java
@@ -2,6 +2,9 @@ package com.sourcegraph.semanticdb_javac;
 
 public enum UriScheme {
   DEFAULT,
+  /** @deprecated Use ZINC instead */
+  @Deprecated
   SBT,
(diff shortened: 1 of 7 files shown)


jdorfman:
Thanks @ckipp01! We will get this reviewed ASAP.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/UriScheme.java
```diff
@@ -2,6 +2,9 @@
 
 public enum UriScheme {
   DEFAULT,
+  /** @deprecated Use ZINC instead */
+  @Deprecated
```
olafurpg, line 6: This is an internal API so no need for deprecation

olafurpg (approved):
LGTM 👍🏻 Thank you! 🙏🏻

olafurpg:
https://github.com/sourcegraph/scip-java/releases/tag/v0.8.7

PR #505, opened 2022-09-30
Test that gradlew/mvnw/mill are executable files.

olafurpg:
Previously, we tried executing these files if they were regular files even if they were not executable.

Reported by @gronk-sg 
 
### Test plan

N/A

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
@@ -40,7 +40,9 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
           "gradlew"
       )
     val gradleCommand: String =
-      if (Files.isExecutable(gradleWrapper))
+      if (
+        Files.isRegularFile(gradleWrapper) && Files.isExecutable(gradleWrapper)
+      )
         gradleWrapper.toString
       else
         "gradle"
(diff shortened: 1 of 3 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -40,7 +40,7 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
           "gradlew"
       )
     val gradleCommand: String =
-      if (Files.isRegularFile(gradleWrapper))
+      if (Files.isExecutable(gradleWrapper))
```
Strum355, line 43: Do we wanna check both? 
> on UNIX systems, checking for execute access checks that the Java virtual machine has permission to search the directory in order to access file or subdirectories.
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isExecutable-java.nio.file.Path-

Strum355: This would return true for directories too, or were we only using isRegularFile to test for existence,m

olafurpg: Directories return false for me 🤔 
```
@ java.nio.file.Files.isExecutable(java.nio.file.Paths.get("project"))
res0: Boolean = false
```

Strum355: Interesting, contradicts the javadoc 🤔 id assume macos falls under the Unix umbrella

olafurpg: It doesn't hurt to test both 🤷🏻 updated the PR

Strum355 (approved):

varungandhi-src (approved):

PR #511, opened 2022-10-07
Fix two issues in the getting started instructions.

sethnelson-toast:
This fixes two issues in the current 'getting started' documentation:

1. Coursier is unable to disambiguate the main class when bootstrapping. Updates instructions to be explicit about the main class.
2. `src scip` is not a valid `src` CLI instruction. The correct instruction is `src code-intel upload`. 

### Test plan

This is a documentation change only.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -23,7 +23,7 @@ container.
 
 ```sh
 $ docker run -v $(pwd):/home/gradle sourcegraph/scip-java:latest scip-java index
-$ src scip upload # (optional) upload index to Sourcegraph
+$ src code-intel upload # (optional) upload index to Sourcegraph
 ```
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -23,7 +23,7 @@ container.
 
 ```sh
 $ docker run -v $(pwd):/home/gradle sourcegraph/scip-java:latest scip-java index
-$ src scip upload # (optional) upload index to Sourcegraph
+$ src code-intel upload # (optional) upload index to Sourcegraph
```
olafurpg, line 26: This repo was originally named "lsif-java" and the invalid `src scip` must be from a bad search/replace

olafurpg (approved):
LGTM 👍🏻 Thank you so much!

PR #519, opened 2022-11-10
Update getting-started.md

Marcuse7:
Fix broken repo link

### Test plan

1. Click on corrected link to check if repo available


Strum355:
@olafurpg I'd imagine we want to direct kotlin issues to https://github.com/sourcegraph/lsif-kotlin instead of this repo. Either way the existing link is wrong though. That being said, should we rename the lsif-kotlin repo to semanticdb-kotlin? Would be good to update the [project name here](https://github.com/sourcegraph/lsif-kotlin/blob/main/settings.gradle.kts#L2) too then, the maven coordinates should be unaffected as theyre `com.sourcegraph:semanticdb-kotlinc` anyways.

olafurpg (approved):
LGTM 👍🏻 I'm OK with directing issues at scip-java instead of the Kotlin repo. I'm +1 with renaming the lsif-kotlin repo into semanticdb-kotlin

olafurpg:
@Marcuse7 Thank you for this contribution!

Strum355:
Think the PR just needs to be rebased so that CI passes, and then we should be good to go @Marcuse7

Marcuse7:
Sorry, was about to mess up the history.

PR #527, opened 2023-01-02
fix: typo in args

ckipp01:
Soooo, this was my mistake. I was curious why this wasn't working as expected and it's because we're checking `args` here instead of `arg`. Yay type safety.

### Test plan

Everything should test the same as before.


Strum355 (approved):

lefou:
A release containing this fix would be very nice. Thanks!

Strum355:
release v0.8.10 with this fix, docker image is already release, maven artifact takes a bit more time to propagate

PR #539, opened 2023-03-14
docs: fix SCIP link on getting started page

enriquegh:
When LSIF was renamed to SCIP a link to Microsoft's site was changed and it does not exist.
Changed the URL to point to our SCIP repo but not sure if we want to point to somewhere else.


olafurpg (approved):
LGTM 👍🏻 Thank you! 🙏🏻

PR #543, opened 2023-04-11
Update Maven/SBT, better JDK distro resolution

keynmol:
_Opening this only to ensure I'm not losing my mind because it works differently on CI_


olafurpg (approved):
LGTM 👍🏻 Feel free to merge

PR #549, opened 2023-04-14
WIP: Fix gradle tool to work with Gradle 8.0+

keynmol:
This allows the tests progress further, but something is still missing 

### Test plan

- Add new tests for both Gradle 8.0 and 7.6.1 so that we don't introduce regressions later


scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -130,6 +130,24 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
   private def initScript(toolchains: GradleJavaToolchains, tmp: Path): Path = {
     val agentpath = Embedded.agentJar(tmp)
     val pluginpath = Embedded.semanticdbJar(tmp)
+
+    val javacModuleOptions =
+      if (toolchains.isJavaAtLeast(9))
+        List(
```
olafurpg, line 136: This list is available in `BuildInfo.javacModuleOptions` if you `.map(_.stripPrefix("-J"))`

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -128,16 +128,19 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
...
+    val agentpath = Embedded.agentJar(tmp)
+    val pluginpath = Embedded.semanticdbJar(tmp)
     val executable =
       toolchains.executableJavacPath() match {
         case Some(path) =>
           s"options.forkOptions.executable = '$path'"
         case None =>
-          ""
+          s"""
+          options.annotationProcessorPath = files('$pluginpath')
```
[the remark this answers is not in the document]
keynmol: I will verify, and possibly just append the plugin to existing paths, instead of overriding the entire path.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
keynmol: Nice!

The current problem is that the time at which this check is run is not right - I need to replicate it in the gradle script

keynmol:
Superseded by #565

PR #550, opened 2023-04-18
Close feature gap between LSIF and SCIP for Scala

olafurpg:
Previously, scip-java had custom support for Scala that was only supported by the LSIF writer. This meant that Scala users were recommended to use LSIF instead of SCIP.

Now that Sourcegraph supports native SCIP uploads (and stores SCIP in the database) it's time to close this feature gap so that Scala users can use SCIP instead of LSIF.

### Test plan

See updated snapshot tests.


keynmol:
Assuming this can be closed @olafurpg ?

PR #554, opened 2023-04-19
Add a welcome screen with common tasks

keynmol:
Let me know what you think.

Welcome banner when SBT is loaded with some commands from https://sourcegraph.github.io/scip-java/docs/contributing.html#helpful-commands being given helpful aliases.

These aliases survive wiping of SBT workspace because they're not part of history, but rather part of the build.

If desired, they can also be used in CI, as individual steps for better reporting, i.e. `sbt --client gradleTests`, `sbt --client snapshotTests`, etc.

image

### Test plan

No testing required as the change is not functional.
Eyeballing CI output is enough.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -564,7 +564,7 @@ usefulTasks :=
       "Runs slow snapshot tests. Indexes a corpus of external Java libraries."
     ).alias("slowSnapshotTests"),
     UsefulTask("snapshots/test", "Runs all snapshot tests")
-      .alias("slowSnapshotTests"),
+      .alias("snapshotTests"),
     UsefulTask(
       "buildTools/testOnly tests.Gradle*",
       "Runs tests for Gradle builds"
--- a/build.sbt
+++ b/build.sbt
@@ -524,3 +524,55 @@ lazy val fatjarPackageSettings = List[Def.Setting[_]](
     ).transform(node).head
   }
 )
+
+import sbtwelcome._
+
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -524,3 +524,55 @@ lazy val fatjarPackageSettings = List[Def.Setting[_]](
...
+       | |___/\___|_| .__/      | |\__,_| \_/ \__,_|
+       |            | |        _/ |                 
+       |            |_|       |__/                  
+       |
+       |${version.value}
+       |
+       |${scala.Console.YELLOW}Scala ${scalaVersion
+    .value}${scala.Console.RESET}
+       |
+       |""".stripMargin
```
keynmol, line 546: Lol, this block of code breaks scalafmt's idempotency, so it's going to be always broken :D 

image

@olafurpg I won't risk upgrading scalafmt to a newer version until we don't have any more pressing PRs in flight.

olafurpg: Scalafmt is such a fragile formatter can't imagine what ***** wrote this piece of software 🙈

olafurpg (approved):

build.sbt
keynmol: :P 

Entire Scala community is still looking for that rapscallion! He's going to _get it_ once we find them.

olafurpg: This person is doing a good job hiding in some corner

PR #555, opened 2023-04-19
Fix snapshot tests

olafurpg:
Previously, the `LibrarySnapshotGenerator` was not working correctly making it impossible to update the snapshots with `snapshots/run`. This commit fixes the issue and updates the snapshots to reflect several versions of scip-java updates.

### Test plan

See updated snapshot tests.


scip-java/src/main/scala/com/sourcegraph/scip_java/commands/IndexDependencyCommand.scala
```diff
@@ -70,12 +71,16 @@ final case class IndexDependencyCommand(
                       snapshotCommand
                         .copy(
                           output = snapshotTarget,
+                          cleanup = cleanup,
+                          targetroot = List(indexTarget),
```
olafurpg, line 75: This line fixed the issue. The other diffs are not necessary but make the `--cleanup` flag work correctly, which is helpful for debugging in cases like this.

keynmol (approved):

PR #556, opened 2023-04-19
Move to SCIP snapshots

olafurpg:
Previously, this repo used SemanticDB and LSIF snapshot tests.
    This commit replaces both of those with SCIP snapshot tests.
    The LSIF snapshot tests are no longer needed since the SCIP snapshot
    format captures the relevant details we tried to assert against with the
    LSIF snapshot tests. The SemanticDB tests are not needed since they look
    almost identical to SCIP snapshot excluding the extra details SCIP
    encodes (cross-repo package information and symbol relationships).

### Test plan

See updated SCIP snapshot tests.


scip-java-proto/src/main/protobuf/scip.proto
```diff
@@ -191,6 +191,23 @@ message Relationship {
...
+    //
+    // On the other hand, in languages with single inheritance and some form
+    // of mixins, you can use is_definition to relate the symbol to the
+    // matching symbol in ancestor classes, and is_reference to relate the
+    // symbol to the matching symbol in mixins.
+    //
+    // NOTE: At the moment, due to limitations of the SCIP to LSIF conversion,
+    // only global symbols in an index are allowed to use is_definition.
+    // The relationship may not get recorded if either symbol is local.
+    bool is_definition = 5;
```
olafurpg, line 210: This isn't strictly needed for this PR but it will be needed in the next PR and having it here allows `ScipPrinters` to support it off the bat.

keynmol (approved):

PR #557, opened 2023-04-19
Close SCIP/LSIF feature gap

olafurpg:
Previously, scip-java had custom support for Scala that was only
    supported by the LSIF writer. This meant that Scala users were
    recommended to use LSIF instead of SCIP.

  Now that Sourcegraph supports native SCIP uploads (and stores SCIP in
    the database) it's time to close this feature gap so that Scala
   users can use SCIP instead of LSIF.

### Test plan
See updated snapshot tests.

### DIFF
--- a/scip-semanticdb/src/main/java/com/sourcegraph/scip_semanticdb/ScipSemanticdb.java
+++ b/scip-semanticdb/src/main/java/com/sourcegraph/scip_semanticdb/ScipSemanticdb.java
@@ -68,6 +68,9 @@ public class ScipSemanticdb {
   }
 
   private String typedSymbol(String symbol, Package pkg) {
+    if (symbol.isEmpty()) {
+      return "";
+    }
     if (symbol.startsWith("local")) {
       return "local " + symbol.substring("local".length());
     }
@@ -91,6 +94,9 @@ public class ScipSemanticdb {
               .collect(Collectors.joining("/"));
       Scip.Document.Builder tdoc = Scip.Document.newBuilder().setRelativePath(relativePath);
       for (SymbolOccurrence occ : doc.sortedSymbolOccurrences()) {
+        if (occ.getSymbol().isEmpty()) {
+          continue;
+        }
         int role = 0;
         if (isDefinitionRole(occ.getRole())) {
           role |= Scip.SymbolRole.Definition_VALUE;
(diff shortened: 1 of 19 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/ScipPrinters.scala
```diff
@@ -0,0 +1,199 @@
...
+            out
+              .append(prefix)
+              .append("documentation ")
+              .append(documentation.replace("\n", "\\n").replace("\t", "\\t"))
+              .append("\n")
+          }
+        0.until(info.getRelationshipsCount)
+          .foreach { n =>
+            val relationship = info.getRelationships(n)
+            out.append(prefix).append("relationship")
```
keynmol, line 181: Should we 

1. Sort relationships by symbol?
2. Sort documentation strings?

I can't reproduce the failures locally, but on CI it seems to be order related:

```diff
+//                      relationship is_reference is_implementation semanticdb maven . . com/airbnb/epoxy/BaseEpoxyTouchCallback#clearView().
 //                      relationship is_reference is_implementation semanticdb maven . . com/airbnb/epoxy/EpoxyTouchHelperCallback#clearView(+1).
 //                      relationship is_reference is_implementation semanticdb maven . . com/airbnb/epoxy/EpoxyTouchHelperCallback#clearView().
-//                      relationship is_reference is_implementation semanticdb maven . . com/airbnb/epoxy/BaseEpoxyTouchCallback#clearView().
 //                      ^ reference semanticdb maven . . com/airbnb/epoxy/EpoxyModelTouchCallback#[T]
```

Here's suggested patch (but this will require regenerating snapshots): 

```suggestion
        val docs =
          0.until(info.getDocumentationCount).map(info.getDocumentation).sorted

        docs.foreach { documentation =>
          out
            .append(prefix)
            .append("documentation ")
            .append(documentation.replace("\n", "\\n").replace("\t", "\\t"))
            .append("\n")
        }

        val relationships = 0
          .until(info.getRelationshipsCount())
          .map(info.getRelationships)
          .sortBy(_.getSymbol())

        relationships.foreach { relationship =>
          out.append(prefix).append("relationship")
```

olafurpg: The failure was probably because I hadn't rebased on top of the scip-snapshots branch, which was failing with the same error. I've rebased now and will see if CI becomes green.

PR #559, opened 2023-04-21
Publish CLI locally and use that version in the sbt-sourcegraph run

keynmol:
1. Dogfood the auto-indexing on this codebase by publishing it locally and using that version to index scip-java itself
2. Do (1) on all PRs to avoid regressions

### Test plan

- After merge to main, verify that the sourcegraph workflow runs successfully and uploads new data
- On this PR, verify that index.scip file is uploaded to the job results

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -21,8 +21,10 @@ jobs:
       - name: Install src
         run: yarn global add @sourcegraph/src
 
-      - name: sbt sourcegraphUpload
-        run: sbt sourcegraphUpload
+      - name: Publish CLI locally
+        run: sbt publishLocal dumpScipJavaVersion
 
+      - name: Upload sourcegraph data
+        run: sbt -Dscip-java-version=$(cat VERSION) sourcegraphUpload
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


olafurpg (approved):

.github/workflows/sourcegraph.yml
```diff
@@ -18,11 +18,25 @@ jobs:
           java-version: 8
           cache: 'sbt'
 
+      - name: Publish CLI locally
+        run: sbt publishLocal dumpScipJavaVersion
+
+      - name: Produce SCIP data of scip-java repository
+        run: sbt -Dscip-java-version=$(cat VERSION) sourcegraphCompile
```
olafurpg, line 25: An alternative to `-Dscip-java-version` is to `set every sourcegraphScipJavaVersion := version.value` inside a command. We can add a command in the root of build.sbt that does something like this

```
commands +=
  Command.command("indexMinimized") { s =>
    "cli/publishLocal" ::
      "set every sourcegraphScipJavaVersion := version.value" ::
      "sourcegraphCompile" :: s
  }
```

I haven't tested this and we can do it separately from this PR. It would be convenient to have a command in the sbt shell to generate an index for local debugging.

keynmol: Yeah, I realised the same later.

Once we confirm that correct scip is being uploaded, I'll go back and make this nicer.

PR #561, opened 2023-04-24
Add `is_reference` relationships between synthetic case class symbols

olafurpg:
Previously, doing "Find references" on a synthetic case class symbol (for example, named parameter of a `copy` method) then it only showed usages of that symbol and the non-synthetic symbol (for example, the
    case class field that matches the `copy` parameter). It didn't
return usages of other synthetic symbols that also relate to the case class field (for example, the `apply` parameter).

This PR fixes the issue by adding `is_reference` relationships between all synthetic symbols that relate to the same non-synthetic symbol.

### Test plan

See updated snapshots.

### DIFF
--- a/.github/workflows/sourcegraph.yml
+++ b/.github/workflows/sourcegraph.yml
@@ -17,9 +17,9 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-          distribution: 'temurin'
+          distribution: "temurin"
           java-version: 8
-          cache: 'sbt'
+          cache: "sbt"
 
       - name: Publish CLI locally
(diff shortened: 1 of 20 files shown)


.github/workflows/sourcegraph.yml
```diff
@@ -17,9 +17,9 @@ jobs:
 
       - uses: actions/setup-java@v3
         with:
-          distribution: 'temurin'
+          distribution: "temurin"
           java-version: 8
-          cache: 'sbt'
+          cache: "sbt"
```
keynmol, line 23: Rest of the changes aside, let's bikeshed about this for a few days.

keynmol (approved):

PR #565, opened 2023-04-25
Replace Gradle init scripts with a proper Gradle plugin

keynmol:
This PR replicates the functionality of the [init script injection](https://sourcegraph.com/github.com/sourcegraph/scip-java/-/blob/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala?L153-275) with a proper Gradle plugin + a much smaller init script.

Benefits:

1. You can edit the plugin code in IDE - it's a normal SBT module
2. Cleaner and doesn't rely as much on values injected as strings
3. More flexible as it can be added to the builds as both an assembly jar (for auto-indexing) or a maven central dependency (in the future)
4. Works for gradle 6, 7, and 8, toolchains included (tested on jdk 8, 11, 17)
5. Most of the logic is done inside the plugin, not in combination of scip-java and init script

Additionally:
- Run CI for Java 11
- Run tests for 3 major gradle versions
- Produce a meaningful message if we get an exception from javac

<details>
<summary>Old TODO list</summary>
TODO: 
- [x] Restore kotlin multi-platform support
- [x] Add dependency dump task
- [x] Modify tests to run on various gradle versions
- [x] Disable publishing of the gradle plugin for _now_
  Currently it's written in Scala as it allows for much faster prototyping. With it not being published but rather injected via assembly, it doesn't matter what language it's written in.

  Once we have this merged and tested in the wild, we can start the process of rewriting the plugin in Java (for maximum compatibility) and publish it - after that, the docs can be modified and all manual steps removed.
</details>

<details>
<summary>Old description from where I was just a young pup helplessly thrashing about</summary>
TODO:
- [x] Kotlin support 
- [x] Rewrite init script
  > One main limitation of init scripts is that they cannot access classes in the buildSrc project (see [Using buildSrc to extract imperative logic](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) for details of this feature).
  

  **Current plan:**
  - Make a proper gradle plugin that does all the steps that Manual Configuration for Gradle performs
- [x] Inject plugin from `GradleBuildTool`
- [x] Convert to SBT
  See https://github.com/scalacenter/gradle-bloop/blob/main/build.sbt as an example of publishing a Gradle plugin from SBT
</details>

### Test plan

- [x] Extra tests for Gradle versions: 6.7 (introduced toolchains), 7.5.1 (latest in 7.x lineage), 8.1.1 (latest)
- [x] Confirm all old tests are passing


semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+
+      val javacPluginVersion = BuildInfo.version
+
+      val javacPluginJar = extraProperties
+        .get("javacPluginJar")
+        .map(_.asInstanceOf[String])
+
+      val javacDep = javacPluginJar
+        .map[Object](jar => project.files(jar))
+        .getOrElse(s"com.sourcegraph:semanticdb-javac:${javacPluginVersion}")
```
keynmol, line 41: We fallback to javac plugin published to maven if there is no jar specified

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+          try {
+            project.getDependencies().add("compileOnly", javacDep)
+            project.getDependencies().add("testCompileOnly", javacDep)
+            true
+          } catch {
+            case exc: Exception =>
+              System
+                .err
+                .println(
+                  s"Failed to add compiler plugin to javac, will go through the agent route: ${exc.getMessage()}"
```
keynmol, line 65: At the time when the plugin is loaded, `compileOnly` configuration might already be resolved, if triggered by something in the build: https://sourcegraph.com/github.com/awslabs/smithy-language-server/-/blob/build.gradle?L212

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+          .configureEach { task =>
+            if (gradle.is5 || (gradle.is6 && !gradle.is6_7_plus))
+              println(
+                task
+                  .asInstanceOf[{
+                      def getToolChain(): Any
+                    }
+                  ]
+                  .getToolChain()
+              )
```
keynmol, line 82: TODO: see what we can do about this.

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+            else {
+              val toolchainCompiler = Option(task.getJavaCompiler().getOrNull())
+                .map(_.getMetadata().getLanguageVersion().asInt())
+
+              val host = System
+                .getProperty("java.version")
+                .split("\\.")
+                .headOption
+                .map(_.toInt)
```
keynmol, line 91: only at play where there is no toolchain provided

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -129,152 +91,35 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
...
-          |                return
-          |            }
-          |        }
-          |    }
-          |}
-          |""".stripMargin
+      s"""
+         | initscript {
+         |     dependencies{ 
+         |         classpath(files("${gradlePluginPath}"))
```
olafurpg, line 106: TIL very cool 👍🏻

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -129,152 +91,35 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
...
+         | }
+         |
+         | import com.sourcegraph.gradle.semanticdb.SemanticdbGradlePlugin
+         |
+         | allprojects {
+         |   project.extra["semanticdbTarget"] = "$targetroot"
+         |   project.extra["javacPluginJar"] = "$pluginpath"
+         |   project.extra["dependenciesOut"] = "$dependenciesPath"
+         |   project.extra["javacAgentPath"] = "$agentpath"
+         |   apply<SemanticdbGradlePlugin>()
```
olafurpg, line 117: 🥳 🥳 😍

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
olafurpg: Could have been a decent inline comment instead of PR comment

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+                .getFiles()
+                .asScala
+                .toList
+                .head
+
+            val args = List(
+              s"-Xplugin:$semanticdbScalac",
+              s"-P:semanticdb:sourceroot:$sourceRoot",
+              s"-P:semanticdb:targetroot:$targetRoot",
+              s"-P:semanticdb:exclude:(src/play/twirl|src/play/routes|src/${System.currentTimeMillis()})", // Ignore autogenerated Playframework files
```
olafurpg, line 202: Nice 👍🏻

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+              // I we actually refer to KotlinCompile at _any_ point here, then
+              // plugin fails with NoClassDefFoundError - because the plugin
+              // classpath is murky
+              //
+              // We also don't want to bundle kotlin plugin with this one as it
+              // can cause all sorts of troubles.
+              //
+              // Instead, we commit the sins of reflection for our limited
+              // needs.
+              val compilerArgs = task
```
olafurpg, line 264: this is fine

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/GradleBuildTool.scala
```diff
@@ -49,70 +48,33 @@ class GradleBuildTool(index: IndexCommand) extends BuildTool("Gradle", index) {
...
-          if (toolchains.isKotlinEnabled)
-            Some("compileTestKotlin")
-          else
-            None,
-          if (toolchains.isKotlinMultiplatformEnabled)
-            Some("compileTestKotlinJvm")
-          else
-            None
-        ).flatten
+        List("clean", "scipCompileAll", "scipPrintDependencies")
```
olafurpg, line 70: What happens with `scipPrintDependencies` if `scipCompileAll` partially fails? Would it make sense to run these commands in reverse order?

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+          false
+      }
+    }
+    def is5 = ver.startsWith("5.")
+  }
+}
+
+class WriteDependencies extends DefaultTask {
+  @TaskAction
+  def printResolvedDependencies() = {
```
olafurpg, line 348: type annotation plz

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -0,0 +1,402 @@
...
+import org.gradle.api.DefaultTask
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier
+import org.gradle.api.tasks.TaskAction
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.api.tasks.scala.ScalaCompile
+
+class SemanticdbGradlePlugin extends Plugin[Project] {
+  override def apply(project: Project) = {
```
olafurpg, line 19: type annotation plz

olafurpg (approved):
This looks really really good 👏🏻 Nice work @keynmol 

I love the plugin approach, it's a much cleaner solution than the unmaintainable init script. I can imagine us doing a lot more with Gradle using this plugin approach.

No blocking comments, just two nitpick suggestions to add type annotations. I'm fine with deleting the init script and moving into this bright new future 🌞

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
keynmol: Given Gradle's compatibility matrix (only 7.3+ supports Java 17) this branch doesn't need anything.

I have a spearate PR coming that will clean this up among other things

PR #566, opened 2023-04-27
Emit inverse relationships

### LINKED ISSUE #413: Missing Scala Precise Code Intelligence when Overriding Fields
`lsif-java` version: `0.7.7`

Type of code: Overriding Fields in Scala

Expected behavior: When clicking on "Find References" on the overridden fields, the abstract field / base field should show up in Precise Code Intelligence. Vice versa.

Actual behavior: It does not show up. Vice versa.

```
trait A {
  val b: Int
}

class B extends A {
  override val b = 10
}
```

olafurpg:
Fixes #413. Previously, "find references" and "find implementations" didn't work as expected with Sourcegraph 5.0 due to https://github.com/sourcegraph/sourcegraph/issues/50927.

This PR is a workaround until that issue gets fixed so that users can index with scip-java and get the best possible navigation experience.

### Test plan

See updated snapshots.

### DIFF
--- a/tests/snapshots/src/main/generated/ujson/IndexedValue.scala
+++ b/tests/snapshots/src/main/generated/ujson/IndexedValue.scala
@@ -34,6 +34,14 @@ sealed trait IndexedValue {
   def index: Int
 //    ^^^^^ definition semanticdb maven maven/com.lihaoyi/ujson_2.13 1.4.0 ujson/IndexedValue#index().
 //          documentation ```scala\ndef index: Int\n```
+//          relationship is_reference is_implementation ujson/IndexedValue.Arr#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.False#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Null#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Num#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.NumRaw#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Obj#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Str#index.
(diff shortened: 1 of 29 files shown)


tests/snapshots/src/main/generated/ujson/IndexedValue.scala
```diff
@@ -34,6 +34,14 @@ sealed trait IndexedValue {
...
 //    ^^^^^ definition semanticdb maven maven/com.lihaoyi/ujson_2.13 1.4.0 ujson/IndexedValue#index().
 //          documentation ```scala\ndef index: Int\n```
+//          relationship is_reference is_implementation ujson/IndexedValue.Arr#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.False#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Null#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Num#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.NumRaw#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Obj#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.Str#index.
+//          relationship is_reference is_implementation ujson/IndexedValue.True#index.
```
keynmol, line 44: Implementation-wise I think the PR is fine, I'm just concerned about plugging the holes in the backend with something that increases the size of the SCIP index.

This example is probably on the extreme end of the spectrum

olafurpg: I'm not so concerned about index size getting bloated from this change. These relationships represent a tiny part of the payload since most of the overhead comes from basic occurrences (just a single `import` statement can emit 5-6 occurrences). Symbols also have a lot of redundancy that gets compressed away.

keynmol (approved):

PR #570, opened 2023-05-02
Fix a regression where the SCIP index contained invalid symbols

olafurpg:
Uploading an index with scip-java v0.8.17 results in the following error in the Sourcegraph UI

```
codeNavSvc.GetImplementations: lsifStore.MonikersByPosition: reached end of symbol while parsing <scheme>, expected a ' ' character
minimized/Issue413Subclass#c.
_____________________________^
```
CleanShot 2023-05-02 at 13 43 46@2x

The problem is that this release introduced a regression where scip-java emitted invalid SCIP symbols. This PR fixes the issue, and adds new checks in the testing infrastructure to prevent this kind of regression from happening again.

### Test plan

See updated snapshots.

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/ScipSymbol.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/ScipSymbol.scala
@@ -1,0 +1,55 @@
+package com.sourcegraph.scip_java
+
+import com.sourcegraph.scip_semanticdb.SymbolDescriptor
+import com.sourcegraph.semanticdb_javac.SemanticdbSymbols
+
+sealed abstract class ScipSymbol {}
+final case class LocalScipSymbol(identifier: String) extends ScipSymbol
+final case class GlobalScipSymbol(
+    scheme: String,
+    packageManager: String,
+    packageName: String,
+    packageVersion: String,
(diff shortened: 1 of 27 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/ScipSymbol.scala
```diff
@@ -0,0 +1,55 @@
...
+              packageVersion,
+              descriptor
+            ) =>
+          GlobalScipSymbol(
+            scheme,
+            packageManager,
+            packageName,
+            packageVersion,
+            parseDescriptors(descriptor)
+          )
```
keynmol, line 36: If we are targeting 2.13, I think you can use direct matching on a string here?

```suggestion
      scipSymbol match {
        case s"$scheme $packageManager $packageName $packageVersion $descriptor" => 
          GlobalScipSymbol(
            scheme,
            packageManager,
            packageName,
            packageVersion,
            parseDescriptors(descriptor)
          )
```

I would check but my local is broken again. It's ridiculous.

keynmol (approved):
Looks good as a stop gap solution, but IMO we should do some light refactoring and separate symbols properly: #573

PR #576, opened 2023-05-04
deps: bump mill-scip to 0.3.2

ckipp01:
This just ensure the version of mill-scip being used has the latest
improvements from semanticdbJava and also this version starts cross
publishing for 0.11.x milestones of Mill.

### Test plan

The existing Mill tests should still be passing. When 0.11.0 is added we can
also add that to the test matrix, but for now I left it as is.


keynmol (approved):
Thanks, Chris :)

PR #581, opened 2023-05-10
Use reflection to work around Gradle's scanning of plugin code

keynmol:
See https://github.com/gradle/gradle/issues/8411#issuecomment-474627356

Additionally, run a basic test on Gradle 5

### Test plan

- New test on Gradle 5 added


olafurpg (approved):
❤️ Scala reflective invocations

PR #596, opened 2023-06-13
feat: add support for 0.11.x series of Mill

ckipp01:
This bumps the version of mill-scip which brings in support for the newly released 0.11.x series of Mill. This also updates the tests to ensure they are running against both the minimal 0.10.x and 0.11.x versions of Mill.

### Test plan

Tests should all be 🍏 .


keynmol:
Thank you!

PR #599, opened 2023-06-27
bzl: add http_archive to bazel-example

jhchabran:
The `http_archive` rule was missing which can lead to confusing errors when the users tries to write the rule themselves and forget about `strip_prefix`, as the `.bazelignore` from this repo will suddenly stop matching the `examples` folder, leading to build errors about the example itself. 

### Test plan

Tested with the example repo from bazel/java docs.

### DIFF
--- a/examples/bazel-example/WORKSPACE
+++ b/examples/bazel-example/WORKSPACE
@@ -22,7 +22,15 @@ local_repository(
     name = "scip_java",
     path = "../.."
 )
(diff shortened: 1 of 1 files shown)


keynmol (approved):
I confirmed that with this change I was able to index a sample Bazel repo (version _had_ to be 5.2.x though)

This also fills the gap in the docs which point to this file which actually didn't contain the required information.

Thank you so much @jhchabran

PR #605, opened 2023-07-05
deps: bump version of mill plugin to 0.3.5

ckipp01:
This also bumps up the default semanticdb version from 4.7.6 to 4.8.1
which brings in some fixes you'll probably want.

### Test plan

Things that be 🟢 should all stay 🟢 .


olafurpg (approved):
Thank you @ckipp01 ! 🙏🏻 

LGTM

PR #610, opened 2023-07-07
Add support to auto-index Bazel builds with an aspect

olafurpg:
Previously, the Bazel example required monkeypatching the `java_library` rule in order to inject the SemanticDB compiler plugin. This PR is an exploration to replace monkeypatching with aspects, which is a Bazel feature that seems made for our exact use-case. The goal of this PR is to make it possible to auto-index Bazel codebases with as little changes as possible to the existing build configuration in WORKSPACE and BUILD files.

With this change, it should be possible to index most Java/Bazel codebases with the following command
```
scip-java index --bazel-scip-java-binary=$(which scip-java)
```

### Test plan

- See new `bazel_aspect` CI job that indexes scip-java itself and the example workspace
- Green CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,12 @@ jobs:
       - run: yarn global add @bazel/bazelisk
       - run: sbt cli/pack
       - run: echo "$PWD/scip-java/target/pack/bin" >> $GITHUB_PATH
-      - run: |
+      - name: Auto-index scip-java codebase
+        run: |
+          scip-java index --build-tool=bazel --bazel-scip-java-binary=$(which scip-java)
+      - run: du -h index.scip
+      - name: Auto-index example/bazel-workspace
+        run: |
           scip-java index --build-tool=bazel --bazel-scip-java-binary=$(which scip-java)
         working-directory: examples/bazel-example
       - run: du -h index.scip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
             docker run -v $PWD/.repos/$REPO:/sources -w /sources sourcegraph/scip-java:latest scip-java index
             file .repos/$REPO/index.scip || (echo "$REPO SCIP index doesn't exist!"; exit 1) 
           }
-          
+
           sudo apt install parallel
           export -f check_repo
 
@@ -72,6 +72,22 @@ jobs:
       - run: bazel build //... --@scip_java//semanticdb-javac:enabled=true
         working-directory: examples/bazel-example
       - run: bazel run @scip_java//scip-semanticdb:bazel -- --sourceroot "$PWD"
+        working-directory: examples/bazel-example
+      - run: du -h index.scip
+        working-directory: examples/bazel-example
+
+  bazel_aspect:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: yarn global add @bazel/bazelisk
+      - run: sbt cli/pack
+      - run: echo "$PWD/scip-java/target/pack/bin" >> $GITHUB_PATH
+      - run: |
+          scip-java index --build-tool=bazel --bazel-scip-java-binary=$(which scip-java)
+      - run: du -h index.scip
+      - run: |
+          scip-java index --build-tool=bazel --bazel-scip-java-binary=$(which scip-java)
         working-directory: examples/bazel-example
       - run: du -h index.scip
         working-directory: examples/bazel-example
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,76 +5,76 @@ on:
       - main
   pull_request:
 jobs:
-  test:
-    runs-on: ${{ matrix.os }}
-    name: Tests
-    strategy:
-      fail-fast: false
-      matrix:
-        # NOTE(olafurpg) Windows is not enabled because it times out due to reasons I don't understand.
-        # os: [windows-latest, ubuntu-latest]
-        os: [ubuntu-latest]
-        java: [8, 11, 17]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v3
-        with:
-          distribution: "temurin"
-          cache: "sbt"
-          java-version: ${{ matrix.java }}
-      - name: Main project tests
-        run: sbt test
+  # test:
+  #   runs-on: ${{ matrix.os }}
+  #   name: Tests
+  #   strategy:
+  #     fail-fast: false
+  #     matrix:
+  #       # NOTE(olafurpg) Windows is not enabled because it times out due to reasons I don't understand.
+  #       # os: [windows-latest, ubuntu-latest]
+  #       os: [ubuntu-latest]
+  #       java: [8, 11, 17]
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - uses: actions/setup-java@v3
+  #       with:
+  #         distribution: "temurin"
+  #         cache: "sbt"
+  #         java-version: ${{ matrix.java }}
+  #     - name: Main project tests
+  #       run: sbt test
 
-  docker_test:
-    runs-on: ${{ matrix.os }}
-    name: Docker CLI tests
-    strategy:
-      fail-fast: true
-      matrix:
-        os: [ubuntu-latest]
-    steps:
-      - uses: actions/checkout@v2
+  # docker_test:
+  #   runs-on: ${{ matrix.os }}
+  #   name: Docker CLI tests
+  #   strategy:
+  #     fail-fast: true
+  #     matrix:
+  #       os: [ubuntu-latest]
+  #   steps:
+  #     - uses: actions/checkout@v2
 
-      - uses: actions/setup-java@v3
-        with:
-          distribution: "temurin"
-          cache: "sbt"
-          java-version: 17
+  #     - uses: actions/setup-java@v3
+  #       with:
+  #         distribution: "temurin"
+  #         cache: "sbt"
+  #         java-version: 17
 
-      - name: Build Dockerised CLI
-        run: sbt cli/docker
+  #     - name: Build Dockerised CLI
+  #       run: sbt cli/docker
 
-      - run: |
+  #     - run: |
 
-          set -eu
-          check_repo() {
-            REPO=$1
-            mkdir -p .repos/$REPO
-            git clone https://github.com/$REPO.git .repos/$REPO
+  #         set -eu
+  #         check_repo() {
+  #           REPO=$1
+  #           mkdir -p .repos/$REPO
+  #           git clone https://github.com/$REPO.git .repos/$REPO
 
-            docker run -v $PWD/.repos/$REPO:/sources -w /sources sourcegraph/scip-java:latest scip-java index
-            file .repos/$REPO/index.scip || (echo "$REPO SCIP index doesn't exist!"; exit 1) 
-          }
+  #           docker run -v $PWD/.repos/$REPO:/sources -w /sources sourcegraph/scip-java:latest scip-java index
+  #           file .repos/$REPO/index.scip || (echo "$REPO SCIP index doesn't exist!"; exit 1)
+  #         }
 
-          sudo apt install parallel
-          export -f check_repo
+  #         sudo apt install parallel
+  #         export -f check_repo
 
-          parallel -j4 check_repo ::: circe/circe indeedeng/iwf-java-sdk
+  #         parallel -j4 check_repo ::: circe/circe indeedeng/iwf-java-sdk
 
-  bazel:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - run: yarn global add @bazel/bazelisk
-      - run: bazel build //... --//semanticdb-javac:enabled=true
-      - run: bazel run scip-semanticdb:bazel -- --sourceroot "$PWD"
-      - run: du -h index.scip
-      - run: bazel build //... --@scip_java//semanticdb-javac:enabled=true
-        working-directory: examples/bazel-example
-      - run: bazel run @scip_java//scip-semanticdb:bazel -- --sourceroot "$PWD"
-        working-directory: examples/bazel-example
-      - run: du -h index.scip
-        working-directory: examples/bazel-example
+  # bazel:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - uses: actions/checkout@v2
+  #     - run: yarn global add @bazel/bazelisk
+  #     - run: bazel build //... --//semanticdb-javac:enabled=true
+  #     - run: bazel run scip-semanticdb:bazel -- --sourceroot "$PWD"
+  #     - run: du -h index.scip
+  #     - run: bazel build //... --@scip_java//semanticdb-javac:enabled=true
+  #       working-directory: examples/bazel-example
+  #     - run: bazel run @scip_java//scip-semanticdb:bazel -- --sourceroot "$PWD"
+  #       working-directory: examples/bazel-example
+  #     - run: du -h index.scip
+  #       working-directory: examples/bazel-example
 
   bazel_aspect:
     runs-on: ubuntu-latest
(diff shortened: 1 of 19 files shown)


olafurpg:
I tried to reproduce the bazel_aspect issue inside a Linux container but I got the indexing working. Need more investigation why the CI is failing here 🤔 I'm wondering if it's related to https://github.com/bazelbuild/bazel/issues/12485

I briefly investigated the Maven test failures and confirmed it's a regression from the `main` branch, but I don't know what's causing the regression. It's weird that only Maven is affected, the Gradle tests are passing. I turned on verbose logs with `-X` but it didn't reveal anything obvious causing the failure

.github/workflows/ci.yml
```diff
@@ -76,6 +76,24 @@ jobs:
...
         working-directory: examples/bazel-example
 
+  bazel_aspect:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: yarn global add @bazel/bazelisk
+      - run: sbt cli/pack
+      - run: echo "$PWD/scip-java/target/pack/bin" >> $GITHUB_PATH
+      - name: Auto-index scip-java codebase
```
keynmol, line 86: IIUC, this part of the docs can be amended now? https://sourcegraph.com/github.com/sourcegraph/scip-java/-/blob/docs/getting-started.md?L332

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
```diff
@@ -657,5 +783,9 @@ object ScipBuildTool {
   // canonical URLs will become 404 links. The lsif-java.json file is not
   // supposed to be written by end-users anyways. It's mostly an implementation
   // default for how we support cross-repo navigation with scip-java.
-  val ConfigFileName = "lsif-java.json"
+  val ConfigFileNames = List("scip-java.json", "lsif-java.json")
+  val isIgnoredAnnotationProcessor = Set(
+    "",
```
keynmol, line 788: IMO this is confusing - let's replace it with `isEmpty` check at use site instead.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -13,15 +26,216 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
-          "requires custom configuration to be added to WORKSPACE and BUILD files. " +
-          "The scip-java repository contains an example WORKSPACE file to demonstrate " +
-          "how to use scip-java with Bazel https://github.com/sourcegraph/scip-java/blob/main/examples/bazel-example/WORKSPACE"
-      )
-    1
+    this.generateAspectFile() match {
+      case None =>
+        1
+      case Some(aspectLabel) =>
+        val buildCommand = ListBuffer.empty[String]
```
keynmol, line 33: nit: I think it's more idiomatic to use `List.newBuilder[String]` if you don't access the full accumulated result _before_ you finish appending to it.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -13,15 +26,216 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
+          .environmentVariables
+          .getOrElse("JAVA_HOME", "")
+        if (javaHome.isEmpty) {
+          index
+            .app
+            .error(
+              "environment variable JAVA_HOME is not set. " +
+                "To fix this problem run `export JAVA_HOME=/path/to/java` and run scip-java index again."
+            )
+          return 1
```
keynmol, line 66: This entire new logic feels very much like it's written in Java, with Scala syntax.

Would be better to 

1. Remove the builder, replacing it with just a regular list, as I think all the places where you append to it are trivially replaced with 
2. May be using a beefed up assert method that throws (and logs the error through `index.app.error`)? There's several early exists in this function that IMO doesn't have to be there

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -13,15 +26,216 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
+              super.visitFile(file, attrs)
+            }
+          }
+        )
+        0
+    }
+  }
+
+  /**
+   * Callback to `os.ProcessOutput.Readlines` that processes the stderr output
```
keynmol, line 160: For this and the various tree walking operations - is there a reason we can't just bring in `os-lib`? Seems like the cli has plenty of dependencies as it is, and it'd be a lot more readable in cases like this to just use `os.walk`

keynmol: Unless I'm misreadingh what you mean by `os.ProcessOutput.Readlines` - this? https://github.com/com-lihaoyi/os-lib/blob/d0661198c915ca037591067ce8967275baca6aae/os/src-jvm/SubProcess.scala#L218

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -13,15 +26,216 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
+            s"Outdated Bazel aspect file found at $aspectPath. To fix this problem, either run again with the flag --bazel-overwrite-aspect-file or update the contents of the file to the following:\n\n$aspectContents\n\n"
+          )
+        return None
+      }
+    } else if (Files.exists(aspectPath)) {
+      index
+        .app
+        .reporter
+        .error(
+          "path $aspectPath already exists and is not a file. To fix this problem, remove this path and try again."
```
keynmol, line 217:
```suggestion
          s"path $aspectPath already exists and is not a file. To fix this problem, remove this path and try again."
```

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
keynmol: Wait, Moped already used os-lib :D 

I'm confused about the use of file visitors then

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
keynmol: As mentioned in https://github.com/sourcegraph/scip-java/pull/610/files#r1262341973 I'm not sure this has to be a mutable thing at all, seems like appending one by one is not strictly necessary in this application.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
```diff
@@ -154,22 +170,23 @@ class ScipBuildTool(index: IndexCommand) extends BuildTool("SCIP", index) {
...
     val javaFiles = allSourceFiles.filter(javaPattern.matches)
     val scalaFiles = allSourceFiles.filter(scalaPattern.matches)
     val kotlinFiles = allSourceFiles.filter(kotlinPattern.matches)
     if (javaFiles.isEmpty && scalaFiles.isEmpty && kotlinFiles.isEmpty) {
-      index
-        .app
-        .warning(
-          s"doing nothing, no files matching pattern '$sourceroot/**.{java,scala,kt}'"
-        )
+      if (config.sourceFiles.isEmpty)
```
keynmol, line 178: Hm, is it intentional that you are not using `allSourceFiles` here?

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
```diff
@@ -478,10 +503,38 @@ class ScipBuildTool(index: IndexCommand) extends BuildTool("SCIP", index) {
...
+      ScipBuildTool.isIgnoredAnnotationProcessor ++
+        index.scipIgnoredAnnotationProcessors
+    val processors = config.processors.filterNot(isIgnoredAnnotationProcessor)
+    if (processors.nonEmpty) {
+      arguments += "-processor"
+      arguments += processors.mkString(",")
+    }
+    arguments ++=
+      fixJavacOptions(config.javacOptions).filterNot(option =>
+        option.startsWith("-Xep") || option.startsWith("-Xplugin:semanticdb") ||
```
keynmol, line 532: Of these three I only know `-Xplugin` - what do `-Xep` and `-XD` prefixes mean?

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
olafurpg: Good catch, this is hacky. I changed the implementation to make the intention explicit: I don't want to report this message when indexing via Bazel because we generate a lot of scip-java.json files that may produce empty indexes, which is fine. I added a field to `Config.reportWarningOnEmptyIndex`, which is `true` by default but `false` when generated via the aspect.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
olafurpg: I moved this logic into `fixJavacOptions` and added comments for the other ones. `-Xep` is for error prone and I'm not sure what `-XD` is for, but it was failing the build for some reason. This might no longer be needed after I updated the aspect to correctly depend on the sibling target, which fixed a bunch of issues (esp. related to the processorpath).

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
keynmol: For the record, the equivalent code (I believe) is

```scala
        val bazelOutLink = Files.readSymbolicLink(bazelOut)
        val out = os.Path(index.finalOutput)

        os.walk(
            os.Path(bazelOutLink),
            skip = path => scipPattern.matches(path.toNIO)
          )
          .foreach { p =>
            os.write.append(out, os.read.stream(p))
          }
```

olafurpg: Old habits die hard. I'll admit that looks very nice, esp. knowing how it uses generators under the hood

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -12,16 +25,225 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
+    if (scipJavaBinary.isEmpty) {
+      index
+        .app
+        .error(
+          "the flag --bazel-scip-java-binary is required to index Bazel codebases. To fix this problem, run scip-java index again with the flag --scip-java-binary=/path/to/scip-java"
+        )
+      return 1
+    }
+
+    val javaHome = index.app.env.environmentVariables.getOrElse("JAVA_HOME", "")
```
keynmol, line 50: It's probably a Mac thing, but 

```
~/p/s/scip-java (olafurpg/aspect) > java -version
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode)
~/p/s/scip-java (olafurpg/aspect) > echo $JAVA_HOME

```

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
```diff
@@ -12,16 +25,225 @@ class BazelBuildTool(index: IndexCommand) extends BuildTool("Bazel", index) {
...
+    Files.createDirectories(index.finalOutput.getParent)
+    val scipPattern = FileSystems.getDefault.getPathMatcher("glob:**.scip")
+    val bazelOut = index.workingDirectory.resolve("bazel-out")
+    if (!Files.exists(bazelOut)) {
+      index
+        .app
+        .error(
+          s"doing nothing, the directory $bazelOut does not exist. " +
+            s"The most likely cause for this problem is that there are no Java targets in this Bazel workspace. " +
+            s"Please report an issue to the scip-java issue tracker if the command `bazel query 'kind(java_*, //...)'` returns non-empty output."
```
keynmol, line 140: _nit:_ don't need interpolators on the last two strings, could make this a stripMargin thing

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
olafurpg: That could be a coursier thing. I use sdkman and it sents JAVA_HOME. It's also configured out of the box on the GH Actions CI runner
```
❯ echo $JAVA_HOME
/Users/olafurpg/.sdkman/candidates/java/current
```

You can fix it by setting the variable with
```
export JAVA_HOME="$(cs java-home)"
```

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
olafurpg: Fixed.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
keynmol: I actually don't use coursier for JDK setup, I use the "standard" way - download a JDK distribution as `.dmg` and install that.

This is a quirk of MacOS which shims `java` executable. Not something that'd happen on a Linux machine.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/BazelBuildTool.scala
olafurpg: I tried running it and hit on
```
java.io.IOException: Is a directory
	at java.base/sun.nio.ch.FileDispatcherImpl.read0(Native Method)
```
because there are many directories with the `.scip` extension, and `os.walk` visits both files and directories. It's possible to `skip = path => os.isFile(path) && ...` but that requires an extra fsstat on every visited file

scip-java/src/main/resources/scip-java/scip_java.bzl
```diff
@@ -0,0 +1,132 @@
...
+
+    if not javac_action:
+        return None
+
+    info = target[JavaInfo]
+    compilation = info.compilation_info
+    annotations = info.annotation_processing
+
+    source_files = []
+    for src in ctx.rule.files.srcs:
```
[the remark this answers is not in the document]
olafurpg: It does
```
❯ cat test/behaviour/reasoner/verification/BUILD
...
native_java_libraries(
    name = "verification",
    srcs = glob(["*.java"]),
...

❯ cat bazel-out/darwin_arm64-fastbuild/bin/test/behaviour/reasoner/verification/verification-linux.scip.json | jq
...
  "sourceFiles": [
    "test/behaviour/reasoner/verification/BoundPattern.java",
    "test/behaviour/reasoner/verification/CompletenessVerifier.java",
    "test/behaviour/reasoner/verification/CorrectnessVerifier.java",
    "test/behaviour/reasoner/verification/ForwardChainingMaterialiser.java",
    "test/behaviour/reasoner/verification/SoundnessVerifier.java"
  ]
}
```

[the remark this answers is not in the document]
olafurpg: From vaticle/typedb repo.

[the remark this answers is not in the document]
keynmol: nice

keynmol (approved):
Excited for this! 

Can't comment much on how idiomatic/correct the starlark script is, but seems fine for what it does.

Very very nice!

PR #611, opened 2023-07-12
Add `-no-relative-path:` flag to control indexing of generated files

olafurpg:
Previously, the SemanticDB compiler plugin errored when indexing auto-generated files outside of the configured `-sourceroot` directory (which is automatically inferred for Bazel builds). This behavior was undesirable because:

- There's no good workaround for the issue
- The error message was cryptic making it difficult to understand what went wrong

For some cases, we were able to detect this situation for Bazel and ignore the indexed file while printing an informative message, but this behavior was also undesirable because we skipping these files means that we can't render hover messages for symbols in those generated files.

This PR fixes the issue by adding a configurable `-no-relative-path:` flag with the following valid options:

- `index_anyways` (default): indexes the file but with no guarantee that it's possible to recover the location of the original generated file. This allows us to display accurate hover tooltips for symbols in these files even if "Go to definition" won't work.
- `skip`: silently ignored these files.
- `warning`: ignore these files and print a message explaining it was skipped.
- `error`: fail the compilation process (old default).

### Test plan

Manually tested this in the example Bazel workspace with the following commands:
```
❯ cd examples/bazel-example
❯ bazel build //... --@scip_java//semanticdb-javac:enabled=true
INFO: Analyzed 4 targets (0 packages loaded, 0 targets configured).
INFO: Found 4 targets...
INFO: Elapsed time: 0.102s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
❯ find bazel-bin/ -type f -name '*.semanticdb'
bazel-bin//src/main/java/example/example.semanticdb/META-INF/semanticdb/src/main/java/example/Example.java.semanticdb
bazel-bin//src/main/java/example/_javac/example/libexample_classes/META-INF/semanticdb/src/main/java/example/Example.java.semanticdb
bazel-bin//src/main/java/source-generator-example/animal.semanticdb/META-INF/semanticdb/src/main/java/source-generator-example/Animal.java.semanticdb
bazel-bin//src/main/java/source-generator-example/_javac/animal/libanimal_classes/META-INF/semanticdb/bazel-out/darwin_arm64-fastbuild/bin/src/main/java/source-generator-example/_javac/animal/libanimal_sources/AutoValue_Animal.java.semanticdb
bazel-bin//src/main/java/source-generator-example/_javac/animal/libanimal_classes/META-INF/semanticdb/src/main/java/source-generator-example/Animal.java.semanticdb
```

Observe that the generated file does not use the `no-relative-path/NUMBER.FILENAME.semanticdb` logic in this PR. To be sure, I manually tested that code path by changing the implementation a bit and confirmed that it generates valid relative paths under `META-INF/semanticdb/no-relative-path/...`.

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
@@ -6,7 +6,6 @@ import com.sun.source.util.TaskListener;
 import com.sun.source.util.Trees;
 import com.sun.tools.javac.model.JavacTypes;
 
-import javax.tools.Diagnostic;
 import javax.tools.JavaFileObject;
 import java.io.IOException;
 import java.net.URI;
@@ -25,7 +24,7 @@ public final class SemanticdbTaskListener implements TaskListener {
   private final SemanticdbReporter reporter;
   private final JavacTypes javacTypes;
   private final Trees trees;
-  private boolean sourceGeneratorsMessageIsLogged = false;
+  private int noRelativePathCounter = 0;
 
   public SemanticdbTaskListener(
       SemanticdbJavacOptions options,
@@ -130,7 +129,11 @@ public final class SemanticdbTaskListener implements TaskListener {
           new String[] {"SimpleFileObject[", "DirectoryFileObject["};
       for (String pattern : knownBazelToStringPatterns) {
         if (toString.startsWith(pattern) && toString.endsWith("]")) {
-          return Paths.get(toString.substring(pattern.length(), toString.length() - 1));
+          Path path = Paths.get(toString.substring(pattern.length(), toString.length() - 1));
+          if (path.isAbsolute()) {
+            return path;
+          }
+          return options.sourceroot.resolve(path);
         }
       }
       throw new IllegalArgumentException("unsupported source file: " + toString);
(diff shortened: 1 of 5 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -190,32 +193,41 @@ private Result<Path, String> semanticdbOutputPath(SemanticdbJavacOptions options
               .resolve(relativePath)
               .resolveSibling(filename);
       return Result.ok(semanticdbOutputPath);
-    } else {
-
-      if (options.uriScheme == UriScheme.BAZEL && options.generatedTargetRoot != null) {
-        try {
-          if (absolutePath.toRealPath().startsWith(options.generatedTargetRoot)) {
```
olafurpg, line 197: @keynmol FYI I kept `options.generatedTargetRoot` even if it's no longer used.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -190,32 +193,41 @@ private Result<Path, String> semanticdbOutputPath(SemanticdbJavacOptions options
...
-              options.sourceroot, absolutePath));
+    switch (options.noRelativePath) {
+      case INDEX_ANYWAY:
+        // Come up with a unique relative path for this file even if it's not under the sourceroot.
+        // By indexing auto-generated files, we collect SymbolInformation for auto-generated symbol,
+        // which results in more useful hover tooltips in the editor.
+        // In the future, we may want to additionally embed the full text contents of these files
+        // so that it's possible to browse generated files with precise code navigation.
+        String uniqueFilename =
+            String.format("%d.%s.semanticdb", ++noRelativePathCounter, absolutePath.getFileName());
```
keynmol, line 206: Do we have any guarantees about `javac` being invoked only once per Bazel target?

Otherwise the `noRelativePathCounter` in second invocation will be reset and files can start overwriting each other

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -190,32 +193,41 @@ private Result<Path, String> semanticdbOutputPath(SemanticdbJavacOptions options
...
+                absolutePath, options.sourceroot),
+            e);
+      case SKIP:
+        return null;
+      case ERROR:
+      default:
+        return Result.error(
+            String.format(
+                "Unable to detect the relative path of '%s'. A common reason for this error is that the file is that this file is auto-generated. "
+                    + "To fix this problem, either configure the -sourceroot:PATH flag to be the parent directory of all indexed files, or "
```
keynmol, line 228: Do you think we can change the error message for Bazel specifically to remove the `-sourceroot` part, which we know is not sufficient?

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -190,32 +193,41 @@ private Result<Path, String> semanticdbOutputPath(SemanticdbJavacOptions options
...
+                    + "configure -no-relative-path:VALUE flag to have one of the following values: index_anyway, skip, warning.",
```
keynmol, line 229: Can use `NoRelativePathMode.values()` again to avoid things drifting if you were to change values/add new ones

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
olafurpg: The output of rebuilding the same Bazel target is guaranteed to produce identical results, unless the build is buggy. You have to use different targets if you're cross-building the same set of sources.

keynmol (approved):

olafurpg:
I will address the unresolved comments separately

PR #613, opened 2023-07-17
Add mnemonic for scip-java aspect

olafurpg:
We use mnemonics in the aspect to filter out actions that produce certain kinds of outputs. By providing a mnemonic, we can do the same for the outputs of the scip-java actions (for example, via `bazel aquery`).

### Test plan

n/a


keynmol (approved):
image

PR #616, opened 2023-07-20
Add option --allow-exporting-global-symbols-from-directory-entries

olafurpg:
Previously, scip-java only allowed cross-index (or cross-repo) navigation between symbols that are compiled to classfiles inside jar files. This behavior mixed an intentional design decision with an accidental implementation detail. The intentional design decision was that the ideal cross-repo navigation experience is from repositories to packages. To achieve this behavior, scip-java didn't assign a package name and package version to symbols that are compiled to directory entries on the classpath.  The accidental implementation detail was that Gradle, Maven, and sbt compile to directories while packages only have jar files.

This commit changes the behavior to treat symbols the same regardless if they are compiled to jar files or directories. The accidental implemenation detail was a hack, and this behavior has come up as a surprise in at least twice cases of trying to integrate scip-java with big propriatery codebases. To get the default old behavior, the flag `--allow-exporting-global-symbols-from-directory-entries` can be set to false.

### Test plan

See updated Gradle and Maven tests.

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ClasspathEntry.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ClasspathEntry.scala
@@ -6,6 +6,7 @@ import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
 
+import scala.annotation.tailrec
 import scala.jdk.CollectionConverters._
 
 import com.sourcegraph.scip_semanticdb.MavenPackage
@@ -15,7 +16,7 @@ import com.sourcegraph.scip_semanticdb.MavenPackage
  * "packageInformation" nodes.
  */
 case class ClasspathEntry(
-    jar: Path,
+    entry: Path,
     sources: Option[Path],
     groupId: String,
     artifactId: String,
@@ -23,7 +24,7 @@ case class ClasspathEntry(
 ) {
   def toPackageHubId: String = s"maven:$groupId:$artifactId:$version"
   def toPackageInformation: MavenPackage =
-    new MavenPackage(jar, groupId, artifactId, version)
+    new MavenPackage(entry, groupId, artifactId, version)
 }
 
 object ClasspathEntry {
@@ -39,13 +40,16 @@ object ClasspathEntry {
    * @param targetroot
    *   @return
    */
-  def fromTargetroot(targetroot: Path): List[ClasspathEntry] = {
+  def fromTargetroot(
+      targetroot: Path,
+      sourceroot: Path
+  ): List[ClasspathEntry] = {
     val javacopts = targetroot.resolve("javacopts.txt")
     val dependencies = targetroot.resolve("dependencies.txt")
     if (Files.isRegularFile(dependencies)) {
       fromDependencies(dependencies)
     } else if (Files.isRegularFile(javacopts)) {
-      fromJavacopts(javacopts)
+      fromJavacopts(javacopts, sourceroot)
     } else {
       Nil
     }
@@ -55,7 +59,8 @@ object ClasspathEntry {
    * Parses ClasspathEntry from a "dependencies.txt" file in the targetroot.
    *
    * Every line of the file is a tab separated value with the following columns:
-   * groupId, artifactId, version, path to the jar file.
+   * groupId, artifactId, version, path to the jar file OR classes directory
+   * path.
    */
   private def fromDependencies(dependencies: Path): List[ClasspathEntry] = {
     Files
@@ -63,9 +68,9 @@ object ClasspathEntry {
       .asScala
       .iterator
       .map(_.split("\t"))
-      .collect { case Array(groupId, artifactId, version, jar) =>
+      .collect { case Array(groupId, artifactId, version, entry) =>
         ClasspathEntry(
-          jar = Paths.get(jar),
+          entry = Paths.get(entry),
           sources = None,
           groupId = groupId,
           artifactId = artifactId,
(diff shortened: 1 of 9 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/commands/IndexSemanticdbCommand.scala
```diff
@@ -52,6 +52,13 @@ final case class IndexSemanticdbCommand(
       "If true, don't report an error when no documents have been indexed. " +
         "The resulting SCIP index will silently be empty instead."
     ) allowEmptyIndex: Boolean = false,
+    @Description(
+      "Determines how to index symbols that are compiled to classfiles inside directories. " +
+        "If true, symbols inside directory entries are allowed to be publicly visible outside of the generated SCIP index. " +
+        "If false, symbols inside directory entries are only visible inside the generated SCIP index. " +
+        "The practical consequences of making this flag false is that cross-index (or cross-repository) navigation does not work between " +
+        "Maven->Maven or Gradle->Gradle projects because those build tools compile sources to classfiles izelinside directories."
```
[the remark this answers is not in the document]
olafurpg: wdym izyou izdon't izunderstand izelinside?? o.O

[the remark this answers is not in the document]
olafurpg: Fixed

scip-semanticdb/src/main/java/com/sourcegraph/scip_semanticdb/BazelBuildTool.java
```diff
@@ -65,7 +65,9 @@ public boolean hasErrors() {
             mavenPackages,
             /* buildKind */ "",
             /* emitInverseRelationships */ true,
-            /* allowEmptyIndex */ true);
+            /* allowEmptyIndex */ true,
+            /* indexDirectoryEntries */ false // because Bazel only compiles to jar files.
```
keynmol, line 69: nit (for future refactor): this many boolean params in a row is a bit annoying, and we can't use named parameters with java class

At some point I think we should generate a builder for this options class to make things clearer.

olafurpg: Agreed. Builders are great when using Java.

olafurpg: We can't used named arguments from Scala if the definition is in Java.

olafurpg: The problem with builders is that we can't make the fields final, but it's not a problem here. We could simplify a lot by just providing sensible default values

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -396,8 +400,43 @@ class WriteDependencies extends DefaultTask {
       .foreach(path => java.nio.file.Files.createDirectories(path.getParent()))
 
     val deps = List.newBuilder[String]
+    val project = getProject()
+
+    // List the project itself as a dependency so that we can assign project name/version to symbols that are defined in this project.
+    // The code below is roughly equivalent to the following with Groovy:
+    //   deps += "$publication.groupId $publication.artifactId $publication.version $sourceSets.main.output.classesDirectory"
+    Try(
```
keynmol, line 408: I think we should log out the exception, but not let it propagate. Even though this feature is optional, having some feedback from failures would be good especially when investigating code we don't have access to.

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ClasspathEntry.scala
```diff
@@ -15,15 +16,15 @@ import com.sourcegraph.scip_semanticdb.MavenPackage
  * "packageInformation" nodes.
  */
 case class ClasspathEntry(
-    jar: Path,
+    entry: Path,
```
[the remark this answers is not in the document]
olafurpg: Done.

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -396,8 +400,50 @@ class WriteDependencies extends DefaultTask {
...
+
+    // List the project itself as a dependency so that we can assign project name/version to symbols that are defined in this project.
+    // The code below is roughly equivalent to the following with Groovy:
+    //   deps += "$publication.groupId $publication.artifactId $publication.version $sourceSets.main.output.classesDirectory"
+    try {
+      for {
+        classesDirectory <- project
+          .getExtensions()
+          .getByType(classOf[SourceSetContainer])
+          .getByName("main")
```
keynmol, line 413: I guess to handle test sources here we'd need to modify the artifact id or something, given that the descriptor doesn't have a notion of classifier like maven dependencies have

olafurpg: Is it common to publish test sources? We can support the test scope as long as there is a reliable way to know the artifact ID of the test config

keynmol: I've seen set ups where a library publishes a testkit from its tests, but now that I think about it - it's frequently discouraged and doesn't work with the rest of tooling: https://github.com/scalacenter/bloop/pulls?q=is%3Apr+author%3Akeynmol+is%3Aclosed+test-jar

So let's ignore this for now until there's a clear need.

olafurpg: We can probably support it if there's demand and there's support to extract the right information

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ClasspathEntry.scala
```diff
@@ -81,36 +87,69 @@ object ClasspathEntry {
...
   }
 
+  private def fromClassesDirectory(
+      classesDirectory: Path,
+      sourceroot: Path
+  ): Option[ClasspathEntry] = {
+    @tailrec
+    def loop(dir: Path): Option[ClasspathEntry] = {
+      if (dir == null || !dir.startsWith(sourceroot))
+        return None
```
keynmol, line 122: nit: this could easily be `if (...) None else ... match`

PR #617, opened 2023-07-21
Move internal API to `com.sourcegraph.internal` package

olafurpg:
Previously, it wasn't clear what parts of scip-java were public APIs and which parts where private APIs. For example, the mill-scip plugin was using internal APIs that had binary breaking API changes with every patch release. This commit fixes the problem by moving all internal APIs under the `com.sourcegraph.internal` package.

### Test plan

Green CI.


jupblb:
`scip.proto` is no longer vendored

PR #621, opened 2023-07-26
Get benchmarks running correctly

olafurpg:
Previously, the benchmarks were crashing because they accessed classes that are shaded. This commit fixes the problem by moving a public member of a shaded class to another class that is not shaded.

### Test plan

n/a. We don't test the benchmarks in CI because they are too slow to run.


olafurpg:
> tests.MillBuildToolSuite

These tests failed due to a seeming flake. Will restart once CI is done with other jobs

keynmol (approved):
Confirmed that this fixes benchmarks locally.
Will follow up with a simple CI setup in a separate PR

PR #627, opened 2023-08-01
Allow custom dependency repositories in SCIP build tool

keynmol:
Previously, when resolving dependencies, we used the following repositories by default:

1. Maven Central
2. Extra repositories such as jitpack, clojars, google's maven repo, etc.

This was (I believe) done for convenience.

There was no way to configure repositories, namely to fully override that list to, for example, avoid Coursier ever going through public repos.

This PR does the following:

- Adds `repositories` (type: array of string) field to `lsif-java.json` with the following semantics:
  - If the field is set, this **fully** overrides the list of remote repositories that will be used (ivy2 local will still be added)
  - if the field is not set, we fall back to previous behaviour: central + ivy2local + some extra remote repos
  - the value `"central"` is handled in a special way in the list and will be resolved to maven central
- Adds `--repositories` repeated flag to `index-dependency` CLI command, with the same behaviour as above: `index-dependency --dependency com.google.code.findbugs:jsr305:3.0.2 --repositories central --repositories http://localhost/bla`
- Refactoring: move the resolving logic to a separate class so we can ensure the repositories are shared

### Test plan

New automated tests

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
@@ -62,6 +62,7 @@ import org.jetbrains.kotlin.config.Services
 import os.CommandResult
 import os.ProcessOutput.Readlines
 import os.SubprocessException
+import com.sourcegraph.scip_java.DependenciesResolver
 
 /**
  * A custom build tool that is specifically made for scip-java.
@@ -165,7 +166,8 @@ class ScipBuildTool(index: IndexCommand) extends BuildTool("SCIP", index) {
     val tmp = Files.createTempDirectory("scip-java")
     Files.createDirectories(tmp)
     Files.createDirectories(targetroot)
-    val deps = Dependencies.resolveDependencies(config.dependencies.map(_.repr))
+    val resolver = Dependencies.resolver(config.repositories)
+    val deps = resolver.resolveDependencies(config.dependencies.map(_.repr))
     val sourceroot = index.workingDirectory
     if (!Files.isDirectory(sourceroot)) {
       throw new NoSuchFileException(sourceroot.toString)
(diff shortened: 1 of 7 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/ScipBuildTool.scala
```diff
@@ -773,6 +780,9 @@ class ScipBuildTool(index: IndexCommand) extends BuildTool("SCIP", index) {
       reportWarningOnEmptyIndex: Boolean = true,
       javaHome: Option[String] = None,
       dependencies: List[Dependency] = Nil,
+      // If repositories aren't provided at all, use all the defaults
+      // if repositories are set, we only use the ones provided explicitly
+      repositories: Option[List[String]] = None,
```
olafurpg, line 785: Is the expectation that users manually write this? My thinking so far has been to rely on `COURSIER_REPOSITORIES` (https://get-coursier.io/docs/other-repositories) and `COURSIER_CREDENTIALS`, which can be overridden from executors without modifying the scip-java.json file. I haven't validated this works, however.

scip-java/src/main/scala/com/sourcegraph/scip_java/commands/IndexDependencyCommand.scala
```diff
@@ -24,6 +25,8 @@ final case class IndexDependencyCommand(
     index: IndexCommand = IndexCommand(),
     @Hidden
     snapshotCommand: SnapshotCommand = SnapshotCommand(),
+    @Repeated
+    repositories: Option[List[String]] = None,
```
olafurpg, line 29: nit: I normally use `List[String]` to avoid having two representations for the empty case (`None` and `Some(Nil)` )

tests/buildTools/src/test/scala/tests/ScipBuildToolSuite.scala
```diff
@@ -57,4 +100,22 @@ class ScipBuildToolSuite extends BaseBuildToolSuite {
             |""".stripMargin
     )
   }
+
+  private def hasMention(name: String) =
+    (out: String) => {
+      assert(out.contains(name), s"Expected output to mention [$name]")
```
olafurpg, line 106: FYI, `clue(..)` can be useful to provide helpful assertion failures
```suggestion
      assert(clue(out).contains(clue(name)), s"Expected output to mention [$name]")
```

tests/buildTools/src/test/scala/tests/ScipBuildToolSuite.scala
```diff
@@ -57,4 +100,22 @@ class ScipBuildToolSuite extends BaseBuildToolSuite {
...
+    }
+
+  private def doesntMention(name: String) =
+    (out: String) =>
+      assert(!out.contains(name), s"Expected output to not mention [$name]")
+
+  private def all(errorAsserts: String => Unit*) =
```
olafurpg, line 113: MUnit is desperately missing a helper to assert multiple things at once

keynmol:
Olaf and I chatted about this, and came to a conclusion that repositories are not part of "package identity", if you will.

Instead, COURSIER_CREDENTIALS (PR coming) + CREDENTIALS_REPOSITORIES env variables should be used to tightly control the fetching process.

PR #629, opened 2023-08-07
Allow picking up credentials from environment variables

keynmol:
Seems like this was not an intentional copy paste.

### Test plan

Manual testing, automated seems to be quite hard to setup (but I will try, still)

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/Dependencies.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/Dependencies.scala
@@ -30,7 +30,6 @@ case class Dependencies(
 object Dependencies {
   val empty = Dependencies(Nil, Fetch.Result(), Fetch.Result())
   private val cache: FileCache[Task] = FileCache[Task]()
-    .noCredentials
     .withCachePolicies(List(CachePolicy.LocalOnly, CachePolicy.Update))
     .withTtl(Duration.Inf)
     .withChecksums(Nil)


build.sbt
```diff
@@ -633,3 +639,7 @@ dumpScipJavaVersion := {
 
   IO.write((ThisBuild / baseDirectory).value / "VERSION", versionValue)
 }
+
+//lazy val cachedPackCli = taskKey[File]("")
+
+//cachedPackCli := {}
```
keynmol, line 645: My sould dies a little that entire CI will re-run after removing these lines..

tests/buildTools/src/test/scala/tests/ScipBuildToolSuite.scala
```diff
@@ -4,6 +4,71 @@ import com.sourcegraph.scip_java.{BuildInfo => V}
...
+    val Password = "world"
+
+    val (requests, _) = PasswordProtectedServer(Username, Password).runWith {
+      run =>
+        val env = Map(
+          "COURSIER_REPOSITORIES" -> run.address.toString(),
+          "COURSIER_CREDENTIALS" -> s"localhost $Username:$Password"
+        )
+
+        val tmp = os.temp.dir(prefix = "scip-java")
```
olafurpg, line 22: FYI, we use `FileLayout.fromString(..., root = workingDirectory)` in `BaseBuildToolSuite`, which makes it easier to create big directory structures without multiple `os.write(...)` invocations.

keynmol: I'll punt this to the next PR for sure 👍

PR #630, opened 2023-08-08
Unconditionally add agent in Gradle plugin, add protobuf codegen test

keynmol:
As reported by customer, in some cases Gradle doesn't complain about semanticdb dependency, but doesn't apply it either.

We don't know the case yet, but as a workaround, it's possible to just add agent unconditionally, and apply it only if semanticdb params didn't make it into the javac options

### Test plan

- Additional test
- Existing tests

### DIFF
--- a/semanticdb-agent/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbAgent.java
+++ b/semanticdb-agent/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbAgent.java
@@ -24,20 +24,20 @@ public class SemanticdbAgent {
 
   public static void premain(String agentArgs, Instrumentation inst) {
     // NOTE(olafur): Uncoment below if you want see all the loaded classes.
-    //    PrintStream logger = newLogger();
-    //    inst.addTransformer(
-    //        new ClassFileTransformer() {
-    //          @Override
-    //          public byte[] transform(
-    //              ClassLoader loader,
-    //              String className,
-    //              Class<?> classBeingRedefined,
-    //              ProtectionDomain protectionDomain,
-    //              byte[] classfileBuffer) {
-    //            logger.println(className);
-    //            return classfileBuffer;
-    //          }
-    //        });
+    // PrintStream logger = newLogger();
+    // inst.addTransformer(
+    // new ClassFileTransformer() {
+    // @Override
+    // public byte[] transform(
+    // ClassLoader loader,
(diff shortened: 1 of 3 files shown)


semanticdb-agent/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbAgent.java
```diff
@@ -24,20 +24,20 @@ public class SemanticdbAgent {
...
+    // public byte[] transform(
+    // ClassLoader loader,
+    // String className,
+    // Class<?> classBeingRedefined,
+    // ProtectionDomain protectionDomain,
+    // byte[] classfileBuffer) {
+    // logger.println(className);
+    // return classfileBuffer;
+    // }
+    // });
```
keynmol, line 40: There's some formatting mismatch :-/

semanticdb-agent/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbAgent.java
```diff
@@ -172,6 +173,10 @@ public static void build(
           case "-Xlint":
             break;
           default:
+            if (option.startsWith("-Xplugin:semanticdb")) {
+              semanticdbAlreadyAdded = true;
+              break;
+            }
```
keynmol, line 179: Don't add semanticdb plugin if it's already been applied

olafurpg (approved):

semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
```diff
@@ -169,21 +169,21 @@ class SemanticdbGradlePlugin extends Plugin[Project] {
...
-                    List(
-                      s"-javaagent:$agentpath",
-                      s"-Dsemanticdb.pluginpath=$pluginpath",
-                      s"-Dsemanticdb.sourceroot=$sourceRoot",
-                      s"-Dsemanticdb.targetroot=$targetRoot"
-                    ).asJava
-                  )
-                }
+            }
+            agentJar.foreach { agentpath =>
```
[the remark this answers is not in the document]
keynmol: done-zo in [6bdfed3](https://github.com/sourcegraph/scip-java/pull/630/commits/6bdfed33d9c201da5c24553c3b454689ced0f538)

PR #649, opened 2023-09-21
Correctly move generated scip file in SBT build tool

keynmol:
### Test plan

Existing tests should pass

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
@@ -51,7 +51,7 @@ class SbtBuildTool(index: IndexCommand) extends BuildTool("sbt", index) {
         .resolve("sbt-sourcegraph")
         .resolve("index.scip")
       if (sourcegraphScip.exitCode == 0 && Files.isRegularFile(inputDump)) {
-        val outputDump = index.workingDirectory.resolve("index.scip")
+        val outputDump = index.workingDirectory.resolve(index.output)
         Files.copy(inputDump, outputDump, StandardCopyOption.REPLACE_EXISTING)
         index.app.info(outputDump.toString)
       }


scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
```diff
@@ -51,7 +51,7 @@ class SbtBuildTool(index: IndexCommand) extends BuildTool("sbt", index) {
         .resolve("sbt-sourcegraph")
         .resolve("index.scip")
       if (sourcegraphScip.exitCode == 0 && Files.isRegularFile(inputDump)) {
-        val outputDump = index.workingDirectory.resolve("index.scip")
+        val outputDump = index.workingDirectory.resolve(index.output)
```
olafurpg, line 54: If we haven't guaranteed that `index.output` is a relative path then we should do this instead
```suggestion
        val outputDump = AbsolutePath.of(index.output, index.workingDirectory)
```

olafurpg (approved):

scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
keynmol: Hm, I think `Path.resolve` handles this regardless 
CleanShot 2023-09-22 at 11 20 00

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#resolve(java.nio.file.Path)

> If the other parameter is an [absolute](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#isAbsolute()) path then this method trivially returns other. If other is an empty path then this method trivially returns this path. Otherwise this method considers this path to be a directory and resolves the given path against this path

Is this what you mean?

olafurpg: Huh, I could swear it didn't work this way at some point  in some JDK but I'm not able to reproduce with 8.0.332-zulu 🤔 LGTM I suppose, this should make a ton of code I've written useless

PR #651, opened 2023-09-25
Fix trailing SBT args and add tests

### LINKED ISSUE #650: trailing arguments doesn't work for sbt projects
Hi,

## Actual behavior

I discovered that no matter what I put into trailing arguments it will be ignored when running against sbt project and instead a generic task `sourcegraphScip` will be called that will eventually invoke compilation of the whole project. 

## Why this is an issue

This is particularly bad for project that cross-compile against multiple scala versions and/or platforms (most of the modern scala libraries) as such compilation can exceed available resources on the CI runner. In order to deal with limited resources scala projects often partition CI pipelines based on the scala version and target platform. Then, commands are invoked only against given pairs of `(scalaVersion, targetPlatfor)`.

Besides that, it is also wasteful, as there is no need to compute that kind of information for every `(scalaVersion, targetPlatfor)`. While there could be some specific sources for some of such pairs, afaik sourcegraph platform doesn't allow for submitting semantic data for different targets.

## Expected behavior

I would expect `scip-java` to respect my trailing arguments and forward them to sbt. This would allow me to call `scip-java` as follows:
```
$ cs launch com.sourcegraph:scip-java_2.13:0.9.6 -- index -- compileScoped 2.13 JVM 
```

where `compileScoped` is a custom task that will compile only 2.13 sources for JVM platform.

This would be also inline with the documentation that states:
> TRAILING ARGUMENTS
> Optional. The build command to use to compile all sources. Defaults to a
> build-specific command. For example, the default command for Maven command is
> 'clean verify -DskipTests'.To override the default, pass in the build command
> after a double dash: 'scip-java index -- compile test:compile'

## Alternatives you have considered

The only alternative I can think of would be to use the [manual configuration](https://sourcegraph.github.io/scip-java/docs/manual-configuration.html) but using `scip-java` is more convenient as I don't want to deal with enabling `semanticDB` plugin at runtime. I also don't want to have the `semanticDB` plugin enabled by default as it slows down the compilation as stated in https://sourcegraph.github.io/scip-java/docs/benchmarks.html 

## How to reproduce

1. clone https://github.com/softwaremill/diffx
2. call `cs launch com.sourcegraph:scip-java_2.13:0.9.6 -- index -- compileScoped 2.13 JVM`
3. observe that it will compile against all targets:
  ```
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/js-2.13/classes ...
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/jvm-2.12/classes ...
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/js-2.12/classes ...
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/jvm-2.13/classes ...
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/jvm-3/classes ...
[info] compiling 39 Scala sources to /home/kghost/workspace/diffx/core/target/js-3/classes ...
  ```

keynmol:
Closes #650 

### Test plan
- New test added

### DIFF
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
@@ -43,8 +43,14 @@ class SbtBuildTool(index: IndexCommand) extends BuildTool("sbt", index) {
 
   private def unconditionallyGenerateScip(): Int =
     Using.resource(sourcegraphSbtPluginFile()) { _ =>
+      val buildCommand =
+        if (index.buildCommand.isEmpty)
+          List("sourcegraphScip")
+        else
+          index.buildCommand
+
       val sourcegraphScip = index
-        .process(List("sbt", "sourcegraphEnable", "sourcegraphScip"))
+        .process(List("sbt", "sourcegraphEnable") ++ buildCommand)
       val inputDump = index
         .workingDirectory
         .resolve("target")
--- a/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
+++ b/scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
@@ -43,19 +43,17 @@ class SbtBuildTool(index: IndexCommand) extends BuildTool("sbt", index) {
 
   private def unconditionallyGenerateScip(): Int =
     Using.resource(sourcegraphSbtPluginFile()) { _ =>
-      val buildCommand =
-        if (index.buildCommand.isEmpty)
-          List("sourcegraphScip")
-        else
-          index.buildCommand
+      val buildCommand = index.finalBuildCommand(List("sourcegraphScip"))
 
       val sourcegraphScip = index
         .process(List("sbt", "sourcegraphEnable") ++ buildCommand)
+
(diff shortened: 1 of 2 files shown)


scip-java/src/main/scala/com/sourcegraph/scip_java/buildtools/SbtBuildTool.scala
```diff
@@ -43,8 +43,14 @@ class SbtBuildTool(index: IndexCommand) extends BuildTool("sbt", index) {
 
   private def unconditionallyGenerateScip(): Int =
     Using.resource(sourcegraphSbtPluginFile()) { _ =>
+      val buildCommand =
+        if (index.buildCommand.isEmpty)
+          List("sourcegraphScip")
```
keynmol, line 48: @olafurpg what's the approach to custom build commands? 

`sourcegraphScip` only produces the SCIP file itself, do we leave it as the last command anyways, or expect the user to add it in if they need the SCIP index as well?

olafurpg: To mirror the Maven/Gradle behavior, the user should only provide commands to compile the codebase to generate SemanticDB

varungandhi-src (approved):

PR #677, opened 2024-01-10
Add and populate new SCIP SymbolInformation fields

### LINKED ISSUE #666: Implement new scip `SymbolInformation` fields
sourcegraph/scip/issues/154 added support for a number of new fields in SCIP's `SymbolInformation`:
 * `kind`
 * `display_name`
 * `signature_documentation`
 * `enclosing_symbol` (what was originally proposed as `owner` or `parent` in the linked issue)

On the Java side the sematicdb data already contains the kind (and properties, eg to build an AbstractMethod), but it does not on the Kotlin side.

For both Java and Kotlin:
 * the semantidb data provides display_name
 * the signature_documentation is already the first line of the documentation
 * semanticdb does not specify the owner/enclosing_symbol for locals anymore, although there is already discussion about adding it back in scalameta/scalameta/issues/1540

nicolas-guichard:
Fixes #666 

Note that the result can only be as good as what the SemanticDB data provides, notably:
 * semanticdb-kotlinc does not provide symbol kinds, properties nor enclosing symbol, this is tracked by sourcegraph/scip-kotlin/issues/66
 * semanticdb-scalac seems to report wrong kinds, for instance some data members are reported as methods

### Test plan

This adds the new fields to the snapshot tests and makes sure they pass.
I have obviously not manually reviewed all 13054 insertions and 3893 deletions to tests/snapshots, but what I think is a reasonable subset.


tests/snapshots/src/main/generated/BaseByteRenderer.scala
```diff
@@ -23,39 +23,57 @@ import upickle.core.{ArrVisitor, ObjVisitor}
...
 //                          ^^^^^^^ reference semanticdb maven . . upickle/
 //                                  ^^^^ reference semanticdb maven . . upickle/core/
 //                                       ^^^^^^^ reference semanticdb maven maven/com.lihaoyi/upickle-core_2.13 1.4.0 upickle/core/ByteOps.
 //                                               ^^^^^^ reference semanticdb maven maven/com.lihaoyi/upickle-core_2.13 1.4.0 upickle/core/ByteOps.Output#
                       (out: T,
 //                     ^^^ definition semanticdb maven maven/com.lihaoyi/ujson_2.13 1.4.0 ujson/BaseByteRenderer#out.
-//                         documentation ```scala\nprivate[this] val out: T\n```
+//                         display_name out
+//                         signature_documentation scala private[this] val out: T
+//                         kind Method
```
nicolas-guichard, line 47: This for instance should be a Field rather than a Method, but I believe this is due to a bug in semanticdb-scalac rather than scip-semanticdb.

tests/snapshots/src/main/generated/com/airbnb/epoxy/ActivityRecyclerPool.kt
```diff
@@ -39,8 +39,10 @@ import java.util.ArrayList
 
 internal class ActivityRecyclerPool {
 //             ^^^^^^^^^^^^^^^^^^^^ definition semanticdb maven . . com/airbnb/epoxy/ActivityRecyclerPool#
+//                                  display_name ActivityRecyclerPool
 //                                  documentation ```kt\ninternal final class ActivityRecyclerPool\n```
```
nicolas-guichard, line 43: In kt files the signature documentation remains in `documentation` for now because semanticdb-kotlinc does not provide the structured signature field and adds the signature to the SematicDB `documentation` field already.

nicolas-guichard (commented):
Pointing to some things that I believe should be fixed upstream.

tests/snapshots/src/main/generated/BaseByteRenderer.scala
keynmol: I'm just going to confer with the council of Semanticdb experts to see if this is intentional due to Scalac class encoding, or accidental.

keynmol:
@nicolas-guichard thanks for the PR! I looked at some of the snapshots and the changes look great.

Before I review the implementation, could you please sync this PR with main? I don't see anything controversial in the PR so far so hopefully we can get this merged and release soon.

In the background I will investigate the semanticdb scalac issue you highlighted - but it won't block this PR as I can reproduce it outside of scip-java.

keynmol:
Thanks for the PR! I moved the commits on top of master in separate PR https://github.com/sourcegraph/scip-java/pull/707 along with a couple of fixes.

We will likely need to wait before releasing this, until Sourcegraph product itself can support signature_documentation (given it's removed from `documentation`)

Issue #686, opened 2024-03-26
semanticdb-javac terminates in error "silently", due to a NullPointerException


spangaer:
Platform: VSCode/Metals/Bloop on Windows 10 with Java 8 GraalVM 20

This issue is a propagation of https://github.com/scalameta/metals/issues/6191

It looks like it went sour in:
- https://github.com/sourcegraph/scip-java/compare/v0.9.3...v0.9.4
- https://github.com/sourcegraph/scip-java/pull/618

Still broken against v0.9.9 against which I did my debugging below.

It seems to be triggered by an annotation on a local variable

```java
  private <T> Function<InvocationHandler, T> getConstructor(Class<T> contract) {
    @SuppressWarnings("unchecked")
    Constructor<T> constructor = (Constructor<T>) 
```

Which throws a `NullPointerException` here
```
JavacTrees.getElement(TreePath) (c:\Program Files\Java\graalvm-ce-java8-20.3.3\src\com\sun\tools\javac\api\JavacTrees.java:319)
JavacTrees.getElement(TreePath) (c:\Program Files\Java\graalvm-ce-java8-20.3.3\src\com\sun\tools\javac\api\JavacTrees.java:119)
SemanticdbTrees.annotationBuilder(AnnotationTree) (Unknown Source:87)
SemanticdbTrees.annotations(Tree) (Unknown Source:71)
SemanticdbVisitor.emitSymbolInformation(Element,Tree) (Unknown Source:134)
SemanticdbVisitor.emitSymbolOccurrence(Element,Tree,Name,Semanticdb$SymbolOccurrence$Role,CompilerRange) (Unknown Source:114)
SemanticdbVisitor.resolveVariableTree(VariableTree,TreePath) (Unknown Source:282)
SemanticdbVisitor.resolveNodes() (Unknown Source:209)
SemanticdbVisitor.buildTextDocument(CompilationUnitTree) (Unknown Source:94)
SemanticdbTaskListener.onFinishedAnalyze(TaskEvent) (Unknown Source:96)
SemanticdbTaskListener.finished(TaskEvent) (Unknown Source:63)
ClientCodeWrapper$WrappedTaskListener.finished(TaskEvent) (c:\Program Files\Java\graalvm-ce-java8-20.3.3\src\com\sun\tools\javac\api\ClientCodeWrapper.java:681)
[...]

PR #691, opened 2024-04-05
Cope with missing annotation parameter

Arthurm1:
This is a quick fix for #686

When `@SuppressWarnings("unchecked")` is used on a class or method, the `TreePathScanner` will scan yielding all nodes before scanning the class/method so all those nodes can be used when emitting symbol info e.g. the following nodes for the annotation are all present...
```
@SuppressWarnings(value = "unchecked")
SuppressWarnings
value = "unchecked"
value
"unchecked"
```

But when the annotation is added to a variable then the scanner yields...
```
@SuppressWarnings(value = "unchecked")
SuppressWarnings
"unchecked"
```

then it scans the variable and then the rest of the annotation.  I've no idea why it does it in this order.  Notice the `value` node is missing.

This PR simply stops the link to `value` being added.  It's unused anyway since `value` doesn't appear in the code and therefore can't be linked to. The reason it's being checked is that `javac` implicitly adds it during the scan.

If the code actually contains `value` e.g. `@SuppressWarnings(value = "unchecked")` then the annotation is scanned before the variable and adds the link correctly.  So I don't think this fix should be an issue.

### Test plan

I've added no tests for this.  Should I add the reproducer code in the issue as a new snapshot test?

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
@@ -84,9 +84,11 @@ public class SemanticdbTrees {
         AssignmentTree assign = (AssignmentTree) param;
         ExpressionTree assignValue = assign.getExpression();
         TreePath variableTreePath = nodes.get(assign.getVariable());
-        Element variableSym = trees.getElement(variableTreePath);
-        String symbol = globals.semanticdbSymbol(variableSym, locals);
-        params.add(tree(assignTree(tree(idTree(symbol)), annotationParameter(assignValue))));
+        if (variableTreePath != null) {
+          Element variableSym = trees.getElement(variableTreePath);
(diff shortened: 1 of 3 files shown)


keynmol:
Hey @Arthurm1 the fix looks as what I'd expect, can you please add the reproducer to snapshot tests? I think this place would fit well: https://github.com/sourcegraph/scip-java/tree/main/tests/minimized/src/main/java/minimized

This is the first NPE I've seen in running scip-java across lots of codebases, so a test would be good to prevent regressions.

Arthurm1:
@keynmol  I've added the test supplied in the reproducer.  I've run the snapshot on Windows - hopefully that won't cause issues in matching in the CI

keynmol (approved):

keynmol:
Thanks @Arthurm1!

I'll do a patch release so that Metals can pull it in.

Aamanuel01 (commented):
tests/snapshots/src/main/generated/tests/minimized/src/main/java/minimized/AnnotationsOnParameterizedTypes.java

PR #703, opened 2024-05-07
feat(semanticdsb-scip): map and add SymbolInformation.Kind to scip output

JamyDev:
This change adds the SymbolInformation.Kind property to DocumentSymbols. The following changes were required to make this work:

1. Updated the SCIP proto to the latest version from [sourcegraph/scip](https://github.com/sourcegraph/scip/blob/main/scip.proto) (keeping the package the same to avoid a bunch of refactoring)
2. Added a mapper to go from SemanticDb Kind to Scip Kind and called it during the SymbolInformation builder.

### Test plan

There's no real way to validate the DocumentSymbols using the snapshot tests. Any recommendations for how to test this properly are appreciated!


JamyDev:
I see that this is a duplicate of #677, will close if favor of that one.

PR #707, opened 2024-05-20
new SymbolInformation fields(rebase on main)

keynmol:
Fixes GRAPH-564
This is a rebase and fixup of https://github.com/sourcegraph/scip-java/pull/677.

As it's introducing changes to the fields that Sourcegraph backend relies on, [I updated the backend to handle it](https://github.com/sourcegraph/sourcegraph/pull/62965).

This PR will be merged and released once a new Sourcegraph release is cut. 

Test plan:
- New tests
- Updates to existing snapshot tests


keynmol:
Update: this PR is ready to go in isolation, but because it changes the way description field is rendered, it will have a negative effect on Sourcegraph customers.

Therefore we have a ticket to start using signature_document if it's present, and fallback to description field if it's not.

PR #728, opened 2024-08-01
Allow passing JVM argsto javac in ScipBuildTool

antonsviridov-src:
A canonical usage for this is passing `--add-opens` flags to the _launcher_ of javac to make sure annotation processors work.

To pass these arguments to the launcher, they have to be perfixed with -J – but arguments like this cannot be passed through the argfile that we use for javacOptions (see https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#command-line-argument-files) so we need to pass them to the command itself. 

For that purpose, we add a `jvmOptions` field to scip-java.json config – these options will have `-J` added to them and passed to the `javac` command.

The test to verify this behaviour relies on an [old version of lombok that requires these options](https://github.com/projectlombok/lombok/issues/2681#issuecomment-748616687)

Additionally, a hidden option `--strict-compilation` is added to the CLI, to prevent error recovery: sometimes scip-java can just ignore javac errors and produce semanticdb artifacts regardless. This complicates testing, so I need an escape hatch).

### Test plan

- New test to ScipBuildToolSuite

### DIFF
--- a/scip-java/src/main/resources/scip-java/scip_java.bzl
+++ b/scip-java/src/main/resources/scip-java/scip_java.bzl
@@ -71,11 +71,20 @@ def _scip_java(target, ctx):
         processorpath += [j.path for j in annotations.processor_classpath.to_list()]
         processors = annotations.processor_classnames
 
+    launcher_javac_flags = []
+    compiler_javac_flags = []
+    for value in compilation.javac_options:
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
This makes a lot of sense. LGTM 👍🏻

PR #732, opened 2024-08-05
GRAPH-775 annotation processing fix

### LINKED ISSUE #727: error: semanticdb-javac:
I am getting an error: semanticdb-javac when running scip-java on my code.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project scip-java-test: Compilation failure: Compilation failure: 
[ERROR] /Users/mwang/code/mwang/scip-java-test/src/main/java/com/imc/mwang/Foo.java:[1] error: semanticdb-javac:
[ERROR] /Users/mwang/code/mwang/scip-java-test/src/main/java/com/imc/mwang/Foo.java:[1] error: semanticdb-javac:
[ERROR] -> [Help 1]
```

After some experimentation, I believe I've narrowed it down to two patterns of code, both related to annotations:

```
public interface Foo {
    // Redundant parentheses around the annotation parameter causes this error
    @Nullable(("reason"))
    String a();

    // As does an annotation parameter of negative float. `-1` appears to be fine
    @Bar(-1.5)
    double c();

    // `-1d` also fails.
    @Bar(-1d)
    double c();

    // It fails regardless of whether the parameter is explicitly named
    @Bar(value = -1d)
    double d();
}

@interface Bar {
    double value();
}
```

antonsviridov-src:
Closes #727

Turns out we weren't handling the unary trees at all, and they're allowed to be values in annotation parameters.

This PR adds support for them.

### Test plan

- New snapshot test

### DIFF
--- a/semanticdb-java/src/main/protobuf/semanticdb.proto
+++ b/semanticdb-java/src/main/protobuf/semanticdb.proto
@@ -380,6 +380,7 @@ enum BinaryOperator {
   LESS_THAN_EQUAL = 18;
 }
 
+// https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.15
 message UnaryOperatorTree {
   UnaryOperator op = 1;
   Tree tree = 2;
--- a/semanticdb-java/src/main/protobuf/semanticdb.proto
+++ b/semanticdb-java/src/main/protobuf/semanticdb.proto
@@ -297,6 +297,7 @@ message Tree {
     AnnotationTree annotation_tree = 9;
     AssignTree assign_tree = 10;
     BinaryOperatorTree binop_tree = 11;
+    UnaryOperatorTree unaryop_tree = 12;
     // -- OUT OF SPEC -- //
   }
 }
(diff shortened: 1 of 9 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -83,7 +83,9 @@ public void finished(TaskEvent e) {
   // exception, it just prints the location with an empty message.
   private void reportException(Throwable exception, TaskEvent e) {
     ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    exception.printStackTrace(new PrintWriter(baos));
+    PrintWriter pw = new PrintWriter(baos);
+    exception.printStackTrace(pw);
+    pw.close();
```
antonsviridov-src, line 88: This line is responsible for not seeing the root cause error message.

I've seen this type of empty message multiple times in customer reports:

> [ERROR] /Users/mwang/code/mwang/scip-java-test/src/main/java/com/imc/mwang/Foo.java:[1] error: semanticdb-javac:

🤞 going forward we'll be able to get more meaningful error reports.

semanticdb-java/src/main/protobuf/semanticdb.proto
```diff
@@ -378,6 +379,22 @@ enum BinaryOperator {
   GREATER_THAN_EQUAL = 17;
   LESS_THAN_EQUAL = 18;
 }
+
+message UnaryOperatorTree {
+  UnaryOperator op = 1;
+  Tree tree = 2;
+}
+
+enum UnaryOperator {
```
olafurpg, line 388: Might be worth linking to JLS to document that this is a complete representation for Java 8 https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.15

olafurpg (approved):
Very clean fix. Nice work tracking this down! 👏🏻

PR #733, opened 2024-08-06
Handle parenthesized trees

antonsviridov-src:
Yet another possible tree wasn't handled, so we add support for it.
This time we don't add anything to semanticdb as it's meaningless.

### Test plan

- new snapshot test

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
@@ -20,6 +20,7 @@ import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.NewArrayTree;
 import com.sun.source.tree.AnnotationTree;
+import com.sun.source.tree.ParenthesizedTree;
 
 import java.util.HashMap;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
```diff
@@ -160,6 +161,9 @@ private Semanticdb.Tree annotationParameter(ExpressionTree expr) {
           unaryOpTree(
               semanticdbUnaryOperator(unaryExpr.getKind()),
               annotationParameter(unaryExpr.getExpression())));
+    } else if (expr instanceof ParenthesizedTree) {
+      ParenthesizedTree parenExpr = (ParenthesizedTree) expr;
+      return annotationParameter(parenExpr.getExpression());
```
antonsviridov-src, line 166: @olafurpg can you confirm it's fine to drill into subtree without creating intermediate semanticdb nodes?

tests/minimized/src/main/java/minimized/AnnotationParameters.java
```diff
@@ -1,23 +1,28 @@
...
+	@SuppressWarnings(value = "unchecked")
+	double test2();
 
-   @BarB(!true)
-   double test3();
-}
+	@BarB(!true)
+	double test3();
 
+	@Nullable(("what"))
```
antonsviridov-src, line 26: this is the new parenthesized node

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTrees.java
olafurpg: This is fine. No need to encode parentheses here, we can add it later if necessary

olafurpg (approved):

PR #735, opened 2024-08-07
Maven plugin for dumping dependencies (cross-repo navigation)

antonsviridov-src:
GRAPH-796

This PR adds a maven plugin that can be used to produce `dependencies.txt` file necessary for cross-repo navigation on Sourcegraph.

It can be used in a standalone invocation:

```
$ mvn com.sourcegraph:maven-plugin:<VERSION>:sourcegraphDependencies
```

Or added directly to maven build and run with `sourcegraph:sourcegraphDependencies`

```xml
  <plugin>
      <groupId>com.sourcegraph</groupId>
      <artifactId>maven-plugin</artifactId>
      <version>${scip-java.version}</version>
      <executions>
          <execution>
              <goals>
                  <goal>sourcegraphDependencies</goal>
              </goals>
          </execution>
      </executions>
  </plugin>
```

### Test plan

- New maven project is added, and a new CI workflow that indexes it using the plugin

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -634,13 +634,12 @@ dumpScipJavaVersion := {
   IO.write((ThisBuild / baseDirectory).value / "VERSION", versionValue)
 }
 
-
 lazy val build = taskKey[Unit](
   "Build `scip-java` CLI and place it in the out/bin/scip-java. "
 )
 
 build := {
   val source = (cli / pack).value
-  val newValue = (ThisBuild / baseDirectory).value / "out"
-  IO.copyDirectory(source, newValue) 
+  val destination = (ThisBuild / baseDirectory).value / "out"
+  IO.copyDirectory(source, destination)
 }
--- a/build.sbt
+++ b/build.sbt
@@ -7,8 +7,6 @@ import java.util.Properties
 import scala.collection.mutable.ListBuffer
 import scala.util.control.NoStackTrace
 
-ThisBuild / version := sys.env.get("CI").fold("dev")(_ => version.value)
-
 lazy val V =
   new {
     val protobuf = "3.15.6"
--- a/build.sbt
+++ b/build.sbt
@@ -7,6 +7,8 @@ import java.util.Properties
 import scala.collection.mutable.ListBuffer
 import scala.util.control.NoStackTrace
 
+ThisBuild / version := sys.env.get("CI").fold("dev")(_ => version.value)
+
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -214,6 +215,33 @@ lazy val scip = project
...
+    javaToolchainVersion := "8",
+    javaOnlySettings,
+    libraryDependencies ++=
+      Seq(
+        "org.apache.maven" % "maven-plugin-api" % "3.6.3",
+        "org.apache.maven.plugin-tools" % "maven-plugin-annotations" % "3.6.4" %
+          Provided,
+        "org.apache.maven" % "maven-project" % "2.2.1"
+      ),
+    Compile / resourceGenerators +=
```
antonsviridov-src, line 231: Maven has a special packaging mode for plugins, which seems to be nothing more than a post-processing step that introspects the plugin's pom.xml + all the Mojo files, then writes the special plugin.xml file.

https://maven.apache.org/guides/plugin/guide-java-plugin-development.html

Setting up a Maven build just for that seems to be counterproductive, so we just write the `plugin.xml` by hand using a template. It's obviously not great, but at this point I don't see a better solution, and we don't expect this plugin to change very frequently. 

`plugin.xml` will be force-kept in sync by tests, because Maven derives _a lot_ of information from it, so if it's broken - things won't work.

For local development this allows you to do `mavenPlugin/publishM2` and use that in maven builds.

olafurpg: Makes sense. I love sbt

maven-plugin/src/main/java/com/sourcegraph/maven/DependencyWriterMojo.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+
+    for (Object dep : artifacts) {
+      if (dep instanceof Artifact) {
+        Artifact artifact = (Artifact) dep;
+        if (artifact.getFile() != null) {
+          builder.append(
+              String.format(
+                  "%s\t%s\t%s\t%s\n",
```
olafurpg, line 68: TSV ftw!

olafurpg (approved):
LGTM 👍🏻 Very nice to finally have a Maven plugin. After using Gradle more, I've become fonder of Maven for Java projects (although sbt is still the best JVM build tool, haven't tried Mill much)

PR #738, opened 2024-08-20
Fix: multiple compilation units in a file lead to information loss

antonsviridov-src:
Fixes GRAPH-753

We discovered that javac issues multiple events per file if there are multiple top level definitions that match what javac considers to be a compilation unit - interfaces or classes for example.

These multiple invocations produce slightly different results, and all writing to the same file, meaning the information gets lost.

To fix this, we defensively read the file if it already exists and carefully deduplicate symbols/occurrences/synthetics before writing back.

### Test plan

- New snapshot results


build.sbt
```diff
@@ -545,6 +545,7 @@ lazy val docs = project
   .enablePlugins(DocusaurusPlugin)
 
 lazy val javaOnlySettings = List[Def.Setting[_]](
+  javafmtOnCompile := false,
```
antonsviridov-src, line 548: This was doing my head in – IntelliJ was formatting one way, running SBT session (~snapshots/run) another way, and IJ constantly producing a warning to load filesystem changes.

Formatting on compile is evil.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -44,7 +45,21 @@ public SemanticdbTaskListener(
   }
 
   @Override
-  public void started(TaskEvent e) {}
+  public void started(TaskEvent e) {
+    // Upon first encounter with a file (before any other tasks are run)
+    // we remove the semanticdb file for this source file to ensure
+    // stale data doesn't cause problems
```
antonsviridov-src, line 51: Now that we read previous state, we shouldn't expect the folder to be clean – without doing so I was getting very confusing results when running snapshots.

build.sbt
olafurpg: > Formatting on compile is evil.

Agreed.

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
```diff
@@ -104,13 +122,84 @@ private void onFinishedAnalyze(TaskEvent e) {
...
 
   private void writeSemanticdb(TaskEvent event, Path output, Semanticdb.TextDocument textDocument) {
-    try {
-      byte[] bytes =
-          Semanticdb.TextDocuments.newBuilder().addDocuments(textDocument).build().toByteArray();
-      Files.createDirectories(output.getParent());
-      Files.write(output, bytes);
-    } catch (IOException e) {
-      this.reportException(e, event);
+    if (Files.exists(output)) {
```
olafurpg, line 125: Nit: it would be nice to extract this condition into a separate function so that `writeSemanticdb` can focus on just writing to disk, not doing merging.

olafurpg (approved):
LGTM 👍🏻 Very nice fix!

PR #750, opened 2024-09-11
Support cast tree in annotation parameters

antonsviridov-src:
Closes GRAPH-889

This doesn't seem to be something that people do often, but it's part of JLS so we should handle it.

### Test plan

- New snapshots


olafurpg (approved):

Arthurm1:
Nice - just hit this in [kafka](https://github.com/apache/kafka/blob/d86cb597902d32ce83f27d65b60df6700cb7a61d/core/src/test/java/kafka/server/logger/RuntimeLoggerManagerTest.java#L60)

antonsviridov-src:
@Arthurm1 excellent, I started to worry I won't find an open source project that uses this :D

PR #751, opened 2024-09-12
Only install coursier and sbt in docker container

antonsviridov-src:
By default coursier preinstalls some applications which aren't necessary, and can cause issues with local setup.
We limit that to just coursier itself and sbt

### Test plan

- Should be covered by existing tests

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import java.nio.file.StandardCopyOption
+import java.nio.file.CopyOption
 import sbtdocker.DockerfileBase
 import scala.xml.{Node => XmlNode, NodeSeq => XmlNodeSeq, _}
 import scala.xml.transform.{RewriteRule, RuleTransformer}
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -591,19 +593,11 @@ lazy val fatjarPackageSettings = List[Def.Setting[_]](
       oldStrategy(x)
   },
   (Compile / Keys.`package`) := {
-    val slimJar = (Compile / Keys.`package`).value
-    val fatJar = crossTarget.value / (assembly / assemblyJarName).value
-    val _ = assembly.value
-    IO.copyFile(fatJar, slimJar, CopyOptions().withOverwrite(true))
-    slimJar
+    assembly.value
```
antonsviridov-src, line 596: I squinted hard at this and I feel like copying is not necessary, as the actual path to the jar doesn't matter.

My hunch is that for some reason copying causes issues (https://github.com/sbt/io/issues/371) which result in flaky CI.

Tests all seem green as well, so I doubt there's a change in published artifacts. Will check the snapshot deployment on maven central to confirm jars are still usable.

olafurpg (approved):

PR #773, opened 2025-02-25
Fix annotated types in new class expression

antonsviridov-src:
Previously, `new @Bla String()` crashed the plugin.
Now we handle it explicitly.

### Test plan

- New snapshot

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -362,7 +362,8 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
           // More complex case, where the type is annotated: `new @TypeParameters String()`
           else if (node.getIdentifier().getKind() == com.sun.source.tree.Tree.Kind.ANNOTATED_TYPE) {
             AnnotatedTypeTree annotatedTypeTree = (AnnotatedTypeTree) node.getIdentifier();
-            if (annotatedTypeTree.getUnderlyingType().getKind() == com.sun.source.tree.Tree.Kind.IDENTIFIER) {
+            if (annotatedTypeTree.getUnderlyingType().getKind()
+                == com.sun.source.tree.Tree.Kind.IDENTIFIER) {
               IdentifierTree ident = (IdentifierTree) annotatedTypeTree.getUnderlyingType();
               emitSymbolOccurrence(
                   sym, ident, ident.getName(), Role.REFERENCE, CompilerRange.FROM_TEXT_SEARCH);
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
@@ -360,10 +360,10 @@ public class SemanticdbVisitor extends TreePathScanner<Void, Void> {
                 CompilerRange.FROM_TEXT_SEARCH);
           }
           // More complex case, where the type is annotated: `new @TypeParameters String()`
-          else if (node.getIdentifier().getKind() == com.sun.source.tree.Tree.Kind.ANNOTATED_TYPE) {
+          else if (node.getIdentifier().getKind() == Tree.Kind.ANNOTATED_TYPE) {
             AnnotatedTypeTree annotatedTypeTree = (AnnotatedTypeTree) node.getIdentifier();
-            if (annotatedTypeTree.getUnderlyingType().getKind()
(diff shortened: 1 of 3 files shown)


tests/snapshots/src/main/generated/tests/minimized/src/main/java/minimized/TypeAnnotations.java
```diff
@@ -0,0 +1,59 @@
...
+//              signature_documentation java public abstract int integer()
+//              kind AbstractMethod
+}
+
+class ClassProcessed<@TypeAnnotation T extends Number> {
+//    ^^^^^^^^^^^^^^ definition semanticdb maven . . minimized/ClassProcessed#
+//                   display_name ClassProcessed
+//                   signature_documentation java class ClassProcessed<T extends Number>
+//                   kind Class
+//                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ definition semanticdb maven . . minimized/ClassProcessed#[T]
```
antonsviridov-src, line 36: Note that this bug is caused by a separate issue, and I will fix it separately.

tests/snapshots/src/main/generated/tests/minimized/src/main/java/minimized/TypeAnnotations.java
```diff
@@ -0,0 +1,59 @@
...
+//                                                    display_name T
+//                                                    signature_documentation java T extends Number
+//                                                    kind TypeParameter
+//                    ^^^^^^^^^^^^^^ reference semanticdb maven . . minimized/TypeAnnotation#
+//                                             ^^^^^^ reference semanticdb maven jdk 11 java/lang/Number#
+
+    public ClassProcessed() {
+//         ^^^^^^^^^^^^^^ definition semanticdb maven . . minimized/ClassProcessed#`<init>`().
+//                        display_name <init>
```
antonsviridov-src, line 45: Another bug just spotted 🙃

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
```diff
@@ -349,12 +350,25 @@ private void resolveNewClassTree(NewClassTree node, TreePath treePath) {
...
+                sym,
+                node,
+                identifierSym.getSimpleName(),
+                Role.REFERENCE,
+                CompilerRange.FROM_TEXT_SEARCH);
+          }
+          // More complex case, where the type is annotated: `new @TypeParameters String()`
+          else if (node.getIdentifier().getKind() == com.sun.source.tree.Tree.Kind.ANNOTATED_TYPE) {
+            AnnotatedTypeTree annotatedTypeTree = (AnnotatedTypeTree) node.getIdentifier();
+            if (annotatedTypeTree.getUnderlyingType().getKind()
```
varungandhi-src, line 365: Does this `getUnderlyingType()` call still succeed even if the code fails to type-check? Or does scip-java not attempt to handle code which cannot be type-checked at all? I'm concerned if this will crash the indexer if some code is ill-typed.

tests/minimized/src/main/java/minimized/TypeAnnotations.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE_USE })
+@interface TypeAnnotation {
+    int integer() default 1;
+}
+
+class ClassProcessed<@TypeAnnotation T extends Number> {
```
varungandhi-src, line 11: Can you mention the open issue for the `@TypeAnnotation T` here in a comment.

```
FIXME(issue: GRAPH-NNN): Incorrect source range for definition below.
```

varungandhi-src (approved):

semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbVisitor.java
antonsviridov-src: There isn't anything in the docs about the typechecking, but given this returns ExpressionTree (and not a ref to some type or a type name), I assume it's just a poorly named method

varungandhi-src: > I assume it's just a poorly named method

How much work would it be to check whether the method can return `null` or not?

antonsviridov-src: I've done a cursory check in JDK sources, and it's never checked for nullability, so I can generally assume that it's not null.

But given it's an arbitrary reference in a Java codebase, it can be null for any number of reasons.

I can add a defensive condition, but that's as far as guarantees go

PR #824, opened 2025-10-02
Support running aspect in Bazel Sandbox by removing cwd override

tpathan-imc:
Right now, the Bazel aspect does not support execution in a bazel sandbox (the default behaviour of Bazel on linux). This is due to the `sourceroot` and `targetroot` definitions pointing to locations within the repository itself, rather than locations in the bazel sandbox. If we attempt to run the aspect as-is inside of the bazel sandbox, we hit read only filesystem exceptions (as we are attempting to write to paths that live outside of the sandbox).

When Bazel executes inside of the sandbox, it creates an execroot that only contains the files that have been included as inputs of the action. The structure of the execroot in practice is a mirror of the actual repository, except filtered down to the known input files. The current working directory of actions executed inside of the sandbox is a mirror of the workspace root outside of the sandbox.

For example, with the bazel root of `examples/bazel-example`, and executing the aspect on `examples/bazel-example/src/main/java/example` target, the sandbox would have the source file at location`$SANDBOX/src/main/java/example/Example.java`, with the cwd set to `$SANDBOX`.

Right now, the aspect passes a `--cwd` flag to `scip-java index` when executing inside the bazel action, which changes the working directory to outside of the sandbox, causing the issues mentioned earlier. However, if we remove the cwd override, and allow `scip-java` to execute with the cwd set to `$SANDBOX`, it still works appropriately without issue. The semanticdb and scip files are generated inside of the sandbox, however upon completion of the action, they will be present in `bazel-out/` in the workspace still, and `scip-java` can still pick this up to generate the final `index.scip`. The semanticdb/scip files generated inside of the sandbox still use relative paths to the workspace root, so they are still applicable.

This PR removes the cwd override to support executing the aspect inside of the bazel sandbox.

### Test plan

We've run a patched version of this aspect on our codebase and it still seems to process correctly. Comparing a diff before and after the `index.scip` file for `examples/bazel-example`, the main difference is that before the project root is set to the current working directory (`$REPO_ROOT/examples/bazel-example`), but after it is now a path into a sandbox (that will no longer exist upon completion). Running `scip snapshot` on the resulting index produces no diff of the output -- but, as before, we need to manually specify the project root as the current working directory.

My understanding is that this change in project root is not an issue, as the upload should be clever enough to switch it with the cwd instead (this is equivalent to the situation where index.scip was generated on a different machine)


jupblb (approved):
Thanks for the contribution! Sorry it took so longe to review it. :)

Issue #769, opened 2025-01-27
Error while building index for Java 21 project - plug-in not found: semanticdb


ofeki-neosec:
I'm trying to build a SCIP index for the following project: https://github.com/mrlonis/mrlonis-spring-boot-monorepo.
I'm using the latest Docker image from the repo, as it has support for Java 21.

For some reason, I'm facing this issue:
```
$ docker run -it -v $(pwd):/sources scip-java scip-java index
Using JVM version '21'
Picked up JAVA_TOOL_OPTIONS: -XX:MaxRAMPercentage=80.0 -XX:+UseContainerSupport
Picked up JAVA_TOOL_OPTIONS: -XX:MaxRAMPercentage=80.0 -XX:+UseContainerSupport
openjdk version "21.0.5" 2024-10-15 LTS
OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.5+11 (build 21.0.5+11-LTS, mixed mode, sharing)
Picked up JAVA_TOOL_OPTIONS: -XX:MaxRAMPercentage=80.0 -XX:+UseContainerSupport
$ /sources/gradlew --no-daemon --init-script /tmp/scip-java10608035674360340155/init-script.gradle -Pkotlin.compiler.execution.strategy=in-process -Dsemanticdb.targetroot=/sources/build/semanticdb-targetroot clean scipPrintDependencies scipCompileAll
Picked up JAVA_TOOL_OPTIONS: -XX:MaxRAMPercentage=80.0 -XX:+UseContainerSupport
Downloading https://services.gradle.org/distributions/gradle-8.12-all.zip
.....................10%......................20%......................30%......................40%......................50%......................60%......................70%......................80%......................90%......................100%

Welcome to Gradle 8.12!

Here are the highlights of this release:
 - Enhanced error and warning reporting with the Problems API
 - File-system watching support on Alpine Linux
 - Build and test Swift 6 libraries and apps

For more details see https://docs.gradle.org/8.12/release-notes.html

To honour the JVM settings for this build a single-use Daemon process will be forked. For more on this, please refer to https://docs.gradle.org/8.12/userguide/gradle_daemon.html#sec:disabling_the_daemon in the Gradle documentation.
Daemon will be stopped at the end of the build
> Task :buildSrc:extractPluginRequests UP-TO-DATE
[...]

PR #826, opened 2025-10-07
[Investigation] Handle Annotation Processing for Test Compilation When Other Processors Present

bmarcaur:
Related to #769 

I am similarly experiencing issues when attempting to index particular internal repositories. When running `scip-java index` I am seeing the following error:

```
Execution failed for task ':log-receiver:compileTestJava'.

> Compilation failed; see the compiler output below.                                                                           error: plug-in not found: semanticdb
```

Taking a closer look at the build scan I noticed that all of the `compileJava` tasks were succeeding. I used that information to isolate the `compileTestJava`. I patched in a small change that appended `--debug` to the Gradle invocation launched by SCIP. I was able to use this to print out the arguments being passed to `javac`. The most glaring difference was that my test compilation was using `-processorpath` for an alternative annotation processor (ErrorProne).

Which, according to the `javac` documentation, `-processorpath` ["Specifies where to find annotation processors. If this option is not used, then the class path is searched for processors."](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/javac.html#:~:text=Specifies%20where%20to%20find%20annotation%20processors.%20If%20this%20option%20is%20not%20used%2C%20then%20the%20class%20path%20is%20searched%20for%20processors) Which led me to question whether it just so happened that we were only adding a competing annotation for the test configurations.

Hence my change. Piggybacking on the annotation mutation when it is discovered that the annotation path is present, we now also append the `semanticdb-plugin` to the `testAnnotationProcessor` similar to how it appends to the `compileOnly` and `testCompileOnly` configurations.

### Note

I am still not 100% certain that this is the absolute root cause, but in making this small change I was able to successfully produce a SCIP index for a repo that previously used to fail.

### Test plan
As of right now this is a simple before and after within a failing repository. That repository is private, but I am willing to put a minimal repro together if it will help build confidence.

### DIFF
--- a/semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
+++ b/semanticdb-gradle-plugin/src/main/scala/SemanticdbGradlePlugin.scala
@@ -79,15 +79,14 @@ class SemanticdbGradlePlugin extends Plugin[Project] {
 
         val compilerPluginAdded =
           try {
-            project.getDependencies().add("compileOnly", javacPluginDep)
+            val dependencies = project.getDependencies()
+            dependencies.add("compileOnly", javacPluginDep)
+            dependencies.add("testCompileOnly", javacPluginDep)
 
             if (hasAnnotationPath) {
-              project
-                .getDependencies()
-                .add("annotationProcessor", javacPluginDep)
+              dependencies.add("annotationProcessor", javacPluginDep)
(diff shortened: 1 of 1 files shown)


jupblb:
Thanks for the contribution! I've rebased your changes and completed the fix in https://github.com/scip-code/scip-java/pull/961. :)

PR #862, opened 2026-04-05
fix(semanticdb-javac): report analysis exceptions as WARNING instead of ERROR

### LINKED ISSUE #861: semanticdb-javac: exceptions in reportException() use Kind.ERROR, breaking compilation for projects with partial classpaths (e.g. Spark)
When `semanticdb-javac` encounters an exception during the `ANALYZE` phase (e.g. a `CompletionFailure` for a missing class), it calls `reportException()`, which reports the failure as `Diagnostic.Kind.ERROR`. This causes javac to exit with a non-zero status, **failing the entire compilation** — even though the catch block comment explicitly says *"we don't want to stop the compilation"*.

## Root cause

**`SemanticdbTaskListener.java` lines 80–93** — the intent is correct, the implementation is not:

```java
} catch (Throwable ex) {
    // Catch exceptions because we don't want to stop the compilation even if this
    // plugin has a bug. We report the full stack trace because it's helpful for bug reports.
    Throwable throwable = ex;
    if (e.getSourceFile() != null) {
        throwable = new CompilationUnitException(
            String.valueOf(e.getSourceFile().toUri().toString()), throwable);
    }
    this.reportException(throwable, e);  // ← ends up as Kind.ERROR → build fails
}
```

**`SemanticdbTaskListener.java` lines 102–108** — `reportException` calls `reporter.error()`:

```java
private void reportException(Throwable exception, TaskEvent e) {
    ByteArrayOutputStream baos = new ByteArrayOutputStream();
    PrintWriter pw = new PrintWriter(baos);
    exception.printStackTrace(pw);
    pw.close();
    reporter.error(baos.toString(), e.getCompilationUnit(), e.getCompilationUnit());
    //      ^^^^^ Kind.ERROR → javac exits non-zero → build fails
}
```

**`SemanticdbReporter.java` lines 54–61** — `error()` uses `Kind.ERROR`:

```java
public void error(String message, Tree tree, CompilationUnitTree root) {
    trees.printMessage(
        Diagnostic.Kind.ERROR,   // ← this is what fails the build
        String.format("semanticdb-javac: %s", message), tree, root);
}
```

## Concrete trigger

Projects that use Apache Spark (or any library with anonymous inner classes compiled from Scala) have types like `org.apache.spark.sql.types.DataType$1` on the Scala classpath but **not** on the Java compile classpath. When `semanticdb-javac` tries to resolve these during deep type analysis, it throws:

```
semanticdb-javac: CompilationUnitException: file:/path/to/BoundReflectFunction.java
  Caused by: CompletionFailure: class file for org.apache.spark.sql.types.DataType$1 not found

1 error
FAILURE: Build failed with an exception.
* What went wrong: Execution failed for task ':compileJava'.
```

The build works fine without `semanticdb-javac`. Removing it restores the build.

## Proposed fix

Add a `warning()` method to `SemanticdbReporter` and update `reportException()` to use it instead of `error()`. Intentional error paths (e.g. `-no-relative-path:error` mode) are left unchanged.

```diff
// SemanticdbReporter.java
+  public void warning(String message, Tree tree, CompilationUnitTree root) {
+    trees.printMessage(
+        Diagnostic.Kind.WARNING, String.format("semanticdb-javac: %s", message), tree, root);
+  }
```

```diff
// SemanticdbTaskListener.java — reportException()
-    reporter.error(baos.toString(), e.getCompilationUnit(), e.getCompilationUnit());
+    reporter.warning(baos.toString(), e.getCompilationUnit(), e.getCompilationUnit());
```

Also change `SemanticdbReporter.exception()` (the `Throwable` overload) from `Kind.ERROR` to `Kind.WARNING` for consistency.

The full stack trace is still reported — useful for bug reports — but as a warning. The build succeeds with partial semanticdb output rather than aborting entirely.

rpalcolea:
`reportException()` was calling reporter.error() which uses Diagnostic.Kind.ERROR, causing javac to exit non-zero and fail the build. The surrounding catch block explicitly states 'we don't want to stop the compilation' — this change makes the implementation match the intent.

Adds `SemanticdbReporter.warning()` for plugin-internal failures, and updates `SemanticdbTaskListener.reportException()` to use it. Intentional error paths (e.g. -no-relative-path:error mode) are unchanged.

Fixes builds for projects with partial classpaths (e.g. Apache Spark) where `CompletionFailure` for anonymous inner classes like DataType was breaking compileJava entirely.

Fixes #861

## What we're testing

`SemanticdbTaskListener.reportException()` previously emitted `Diagnostic.Kind.ERROR`,
causing javac to exit non-zero and fail the build when semanticdb-javac hit an internal
exception. After the fix it emits `Kind.WARNING`, so the build succeeds with partial
semanticdb output.

---

## New unit test

`PartialClasspathSuite.scala` in `tests/unit/src/test/scala/tests/`.

### How it triggers `reportException()`

The test pre-creates the semanticdb output path (`META-INF/semanticdb/example`) as a
**regular file** instead of a directory. When `SemanticdbTaskListener.writeSemanticdb()`
calls `Files.createDirectories(output.getParent())`, it throws `FileAlreadyExistsException`
(a subtype of `IOException`). The existing `catch (IOException e)` block catches it and
calls `reportException()` → `reporter.warning()` → `trees.printMessage(Kind.WARNING, ...)`.

This approach is JDK-version-agnostic. An earlier design attempted to trigger
`CompletionFailure` by deleting an inner class file from the classpath, but Java 21+
handles missing inner class files silently (omitting them rather than throwing), so
that approach was not reliable across JDK versions.

### Test 1 — compilation succeeds with a warning

- Blocks the output directory, compiles `UsesA.java`
- Asserts `result.isSuccess == true`
- Asserts `result.stdout.contains("warning:")`
- Asserts `!result.stdout.contains("\nerror:")`

### Test 2 — semanticdb still produced for healthy files

- Compiles `UsesA.java` (triggers exception) alongside `Healthy.java` (clean)
- Asserts `result.isSuccess == true`
- Asserts `Healthy.java` has a semanticdb document in the output

---

## Existing tests that continue to pass

Full unit suite was run on the branch — all tests pass:

```
sbt "unit/test"
```

Key suites that exercise compilation behaviour:
- `OverridesSuite` — type resolution across files
- `JavacClassesDirectorySuite` — targetroot output
- `SnapshotCommandSuite` — end-to-end symbol output

### DIFF
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbReporter.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbReporter.java
@@ -28,7 +28,9 @@ public class SemanticdbReporter {
     e.printStackTrace(writer);
     writer.println(
         "Please report a bug to https://github.com/sourcegraph/semanticdb-java with the stack trace above.");
-    trees.printMessage(Diagnostic.Kind.ERROR, baos.toString(), tree, root);
+    // Use WARNING so that internal exceptions never fail the compilation.
+    // The full stack trace is preserved for bug reports.
+    trees.printMessage(Diagnostic.Kind.WARNING, baos.toString(), tree, root);
   }
 
   public void exception(Throwable e, TaskEvent task) {
@@ -58,5 +60,15 @@ public class SemanticdbReporter {
     // the reporter API so the message goes to stderr instead for now.
     trees.printMessage(
         Diagnostic.Kind.ERROR, String.format("semanticdb-javac: %s", message), tree, root);
+  }
+
+  /**
+   * Reports a warning diagnostic. Use this for internal plugin failures (e.g. exceptions during
+   * analysis) that should not fail the compilation — the build continues with partial semanticdb
+   * output rather than aborting.
+   */
+  public void warning(String message, Tree tree, CompilationUnitTree root) {
+    trees.printMessage(
+        Diagnostic.Kind.WARNING, String.format("semanticdb-javac: %s", message), tree, root);
   }
 }
--- a/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
+++ b/semanticdb-javac/src/main/java/com/sourcegraph/semanticdb_javac/SemanticdbTaskListener.java
@@ -94,17 +94,20 @@ public final class SemanticdbTaskListener implements TaskListener {
     }
   }
 
-  // Uses reporter.error with the full stack trace of the exception instead of
-  // reporter.exception
-  // because reporter.exception doesn't seem to print any meaningful information
-  // about the
-  // exception, it just prints the location with an empty message.
+  // Uses reporter.warning with the full stack trace of the exception instead of
+  // reporter.exception because reporter.exception doesn't seem to print any meaningful
+  // information about the exception, it just prints the location with an empty message.
+  //
+  // WARNING (not ERROR) is intentional: the catch block above says "we don't want to stop
+  // the compilation", but Kind.ERROR was causing javac to exit non-zero and fail the build.
+  // Reporting as a warning preserves the full stack trace for bug reports while allowing
+  // compilation to succeed with partial semanticdb output.
   private void reportException(Throwable exception, TaskEvent e) {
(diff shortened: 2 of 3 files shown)


jupblb (commented):
This fix looks valid. Could you please rebase on top of `origin/main`?

PR #932, opened 2026-06-26
docs: add missing AGENTS.md (and README where missing)

trly:
Adds the missing root `AGENTS.md` for this repository (and a root `README.md` where one was also missing).

Generated by a batch change following the repository-documentation skill.

[_Created by Sourcegraph agentic batch change._](https://sourcegraph.sourcegraph.com/batch-change-agents/253)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,61 @@
+# AGENTS.md
+
+Guidance for coding agents working in `scip-java`, a Java and Kotlin indexer that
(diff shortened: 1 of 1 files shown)


jupblb (commented):
A few things:
1. This repo is going through some major changes, so the contents of this doc may be already outdated.
2. My problem with `AGENTS.md` files is that they can get outdated easily pretty quickly. I don't think there will be anyone to maintain it in the near future. I also believe that this project is small enough for AI agents to be able to function well inside without additional help.

PR #959, opened 2026-07-02
Disable Maven snapshots on main

jupblb:
Stops the Maven Central workflow from publishing snapshots on every main push. Release publishing remains tag-driven.


jupblb:
Closing: we want to keep snapshot publishing on main and fix the Maven Central publish failure instead.

PR #973, opened 2026-07-03
scip-kotlinc Kotlin update, fixes and improvements

nicolas-guichard:
We've been maintaining a semanticdb-kotlinc fork at https://github.com/mozsearch/semanticdb-kotlinc for some time now, which was kept up-to-date with Kotlin updates with a couple of extra fixes and improvements along the way.

Since this repo seems to be actively maintained again, I've rebased our changes on top of the new scip-kotlinc to submit them here.

Now I only just saw https://github.com/scip-code/scip-java/pull/963 which plans to drop the plugin in favor of the Analysis API (which makes sense), so… This may still be useful until that PR is fully fleshed-out.


jupblb:
Hi @nicolas-guichard, thank you for this outstanding contribution! It's true that we experimented a bit with switching to `scip-kotlin-analysis` to make the indexer work for multiple Kotlin versions, but it's early work. I'd be very happy to merge this PR if you fix the CI checks.

nicolas-guichard:
That should fix the CI failures except the test runs against https://github.com/square/okio and https://github.com/hoc081098/FlowExt because they target 2.2.0.

square/okio master has updated to 2.2.21 and hoc081098/FlowExt master still targets 2.2.0.

An option would be to patch them to 2.4.0 in our [GitHub Actions workflow](https://github.com/scip-code/scip-java/blob/main/.github/workflows/repos.yaml) between the “Clone repository” and “Run scip-java index” steps. Hopefully that should just work™ unless they also use private API.

Another option would be to replace them with other projects that target 2.4.0.

What do you prefer, or do you have a better option in mind?

jupblb:
> Another option would be to replace them with other projects that target 2.4.0.

This is definitely a preferred option. We do not have repo-specific tests (we do not check contents of the generated indexes). Please just pick anything that works and is relatively popular/maintained. :)

nicolas-guichard:
The KotlinConf App was failing because it doesn't specify `dependencyResolutionManagement.repositoriesMode` and thus defaults to `RepositoriesMode.PREFER_PROJECT`, but overriding the repositories at the project level caused the build to fail. (Related to #847, #969 only fixed the `FAIL_ON_PROJECT_REPOS` case.)

I've included a fix for this here, but I'm happy to split that into another PR if you prefer.

jupblb (approved):
Thank you! I believe this PR is ok to merge as-is. Awesome work! :)

